PR #43, opened 2023-07-26
Define CAMARA Event using cloudevents in OAS notification

### LINKED ISSUE #8: Provides specific CAMARA OAS for notification event
Prepared with @patrice-conil 
As we are offering “notification event” for a lot of our API we need to clarify the technical assets. As of today, we are defining the notification event in the **same** OAS than the ‘nominal’ API resource itself. It could trigger confusion as it mixes endpoint provided by server from the notification endpoint that must be on client listener side.

Our proposal is

- to provide a separate OAS only describing the notification endpoint. We’ll provided here generic CAMARA event notification structure. This swagger will be available to app developer to generate their listener endpoint.
- To describe event notification API per API in the callback structure (as done in QoS).

patrice-conil:
#### What type of PR is this?

* enhancement/feature

This PR increase interoperability between our ecosystem and rest of world using a defacto standard

Fixes #8 

BREAKING CHANGES:

- Impacts APIs that define Event notifications (QoD, device status, ...)
- Need to update guidelines to integrate cloudevents references.

#### Changelog input

```
Integrates the cloudevents specification as the basis for CAMARA notifications
```

#### Additional documentation

### DIFF
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -190,10 +190,6 @@ components:
         data:
           type: object
           description: Event details payload described in each CAMARA API and referenced by its type
-        subject:
-          description: Describes the subject of the event in the context of the event producer (identified by source).
-          type: string
-          example: ROAMING
         time:
           $ref: "#/components/schemas/DateTime"
 
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -47,8 +47,8 @@ externalDocs:
   description: Product documentation at CAMARA
   url: https://github.com/camaraproject/
 security:
-  - oAuth2ClientCredentials: []
-  - {}
+    - notificationsBearerAuth: []
+    - {}
 
 servers:
   - url: '{apiRoot}'
@@ -114,12 +114,10 @@ paths:
           $ref: "#/components/responses/Generic503"
 components:
   securitySchemes:
-    oAuth2ClientCredentials:
-      type: oauth2
-      flows:
-        clientCredentials:
-          tokenUrl: '{tokenUrl}'
-          scopes: {}
+    notificationsBearerAuth:
+      type: http
+      scheme: bearer
+      bearerFormat: "{$request.body#/webhook/notificationAuthToken}"
   schemas:
     ErrorInfo:
       type: object
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -8,7 +8,7 @@ info:
 
     A lot of CAMARA APIs offer the capability to API consumer to receive events.
     Event data are defined in each API definition but in order to provide consistency across CAMARA APIs and to increase 
-    interoperability we will use [cloudevents](https://cloudevents.io/) specifications. In particular, CAMARA Event will 
+    interoperability we will use [cloudevents](https://cloudevents.io/) specifications. In particular, every CAMARA Event will 
     be defined using [cloudevents-json-format](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/json-format.md)
    
     # Relevant terms and definitions
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -177,8 +177,8 @@ components:
           $ref: '#/components/schemas/Source'
         type:
           type: string
-          description: 'type of event as defined in each CAMARA API (e.g.: org.camara.qod.qos-status-changed-event)'
-          example: org.camara.qod.qos-status-changed-event
+          description: 'type of event as defined in each CAMARA API (e.g.: org.camaraproject.qod.qos-status-changed-event)'
+          example: org.camaraproject.qod.qos-status-changed-event
           minLength: 15
         specversion:
           type: string
@@ -280,7 +280,7 @@ components:
       value:
         id: 123e4567-e89b-12d3-a456-426655440000
         source: com.orange.camara.qod
-        type: org.camara.qod.qos-status-changed-event
+        type: org.camaraproject.qod.qos-status-changed-event
         specversion: '1.0'
         time: 2023-01-17T13:18:23.682Z
         datacontenttype: application/json
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -31,14 +31,14 @@ info:
       * **Data**:  Domain-specific information about the occurrence (i.e. the payload). This might 
     include information about the occurrence, details about the data that was changed, or more.
   
+    # API Functionality
     
-    # API Functionality   
-      
     Only one endpoint/operation is provided: `POST /your_webhook_notification_url`
-    
     This endpoint describes the event notification received on subscription listener side when the event occurred. A detailed description of the event notification is provided in the CAMARA API design guideline document.
   
-  
+    # Security
+    
+    Security is the responsibility of the API provider. The API provider must include its requirements in the security section of this specification or remove this section if it does not support any authentication scheme.
 
   termsOfService: http://swagger.io/terms/
   contact:
@@ -46,7 +46,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.1.0-wip
+  version: 0.1.0
 externalDocs:
   description: Product documentation at CAMARA
   url: https://github.com/camaraproject/
@@ -57,7 +57,7 @@ servers:
   - url: '{apiRoot}'
     variables:
       apiRoot:
-        default: http://localhost:8080
+        default: https://localhost:8080
         description: Can be any notification server address sent by the client application
 tags:
   - name: Cloud Event
@@ -145,7 +145,9 @@ components:
     DateTime:
       type: string
       format: date-time
-      description: Timestamp of when the occurrence happened. Must adhere to RFC 3339.
+      description: |
+        Timestamp of when the occurrence happened. Must adhere to RFC 3339. 
+        WARN: This optional field in CloudEvents specification is required in CAMARA APIs implementation.
       example: '2018-04-05T17:31:00Z'
 
     Source:
@@ -179,7 +181,7 @@ components:
           type: string
           description: 'type of event as defined in each CAMARA API (e.g.: org.camaraproject.qod.qos-status-changed-event)'
           example: org.camaraproject.qod.qos-status-changed-event
-          minLength: 15
+          minLength: 25
         specversion:
           type: string
           description: Version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version)
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -151,13 +151,17 @@ components:
       format: uri-reference
       minLength: 1
       description: "Identifies the context in which an event happened - be a non-empty `URI-reference."
-      example:
-        - https://github.com/cloudevents
-        - mailto:cncf-wg-serverless@lists.cncf.io
-        - urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
-        - cloudevents/spec/pull/123
-        - /sensors/tn-1234567/alerts
-        - 1-555-123-4567
+     examples:
+        absolute_uri_https:
+          value: https://github.com/cloudevents
+        absolute_uri_mailto:
+          value: mailto:cncf-wg-serverless@lists.cncf.io
+        URN_with_UUID:
+          value: urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
+        application_specific_relative_reference:
+          value: /cloudevents/spec/pull/123
+        application_specific:
+          value: 1-555-123-4567
 
     CloudEvent:
       description: The notification callback
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -6,7 +6,7 @@ info:
 
     # Introduction
 
-    A lot of CAMARA APIs offer the capability to consumer to receive event.
+    A lot of CAMARA APIs offer the capability to API consumer to receive events.
     Event data are defined in each API definition but in order to provide consistency across CAMARA APIs and to increase 
     interoperability we will use [cloudevents](https://cloudevents.io/) specifications. In particular, CAMARA Event will 
     be defined using [cloudevents-json-format](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/json-format.md)
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -277,7 +277,7 @@ components:
     QOS_STATUS_CHANGED_EXAMPLE:
       value:
         id: "123e4567-e89b-12d3-a456-426655440000"
-        source: "com.orange.camara.qod"
+        source: "https://notificationSendServer12.supertelco.com"
         type: "org.camaraproject.qod.qos-status-changed-event"
         specversion: "1.0"
         time: "2023-01-17T13:18:23.682Z"
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -67,13 +67,17 @@ tags:
     description: |
       Events received on subscription listener side.
 paths:
-  /cloudevents:
+  /your_webhook_notificationUrl:
     post:
       tags:
         - CAMARA Cloud Event notification endpoint
       summary: "Cloud Event notification endpoint to notify consumer that statement of fact had occurred"
-      description: Cloud Event notification endpoint to notify consumer that statement of fact occurred. This endpoint must be exposed on the consumer side.
-      operationId: notifyCloudEvent
+      description: |
+        INFORMATIVE ENDPOINT: The value of this endpoint is freely declared by each client app by means of resource-based
+        subscription or instance-based subscription. /your_webhook_notificationUrl is 
+        just a convention naming referring to an absolute URL, indeed the one indicated by API client
+        in the triggering of the procedure (resource-based or instance-based).
+      operationId: sendEvent
       requestBody:
         required: true
         content:
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -54,14 +54,11 @@ security:
   - oAuth2ClientCredentials: []
 
 servers:
-  - url: '{apiRoot}/{basePath}'
+  - url: '{apiRoot}'
     variables:
       apiRoot:
         default: http://localhost:8080
-        description: API root
-      basePath:
-        default: /event-notification/v0
-        description: Base path for the event notification API
+        description: Can be any notification server address sent by the client application
 tags:
   - name: Cloud Event
     description: |
@@ -180,8 +177,8 @@ components:
           $ref: '#/components/schemas/Source'
         type:
           type: string
-          description: 'type of event as defined in each CAMARA API (e.g. org.camara.api.qod.QosStatusChangedEvent)'
-          example: org.camara.api.qod.QosStatusChangedEvent
+          description: 'type of event as defined in each CAMARA API (e.g.: org.camara.qod.qos-status-changed-event)'
+          example: org.camara.qod.qos-status-changed-event
           minLength: 15
         specversion:
           type: string
@@ -283,10 +280,9 @@ components:
       value:
         id: 123e4567-e89b-12d3-a456-426655440000
         source: com.orange.camara.qod
-        type: org.camara.api.qod.QosStatusChangedEvent
+        type: org.camara.qod.qos-status-changed-event
         specversion: '1.0'
         time: 2023-01-17T13:18:23.682Z
-        subject: QOD/QOS_STATUS_CHANGED
         datacontenttype: application/json
         data:
           sessionId: 6e8bc430-9c3a-11d9-9669-0800200c9a66
--- a/artifacts/notification-as-cloud-event.yaml
+++ b/artifacts/notification-as-cloud-event.yaml
@@ -1,0 +1,268 @@
+openapi: 3.0.3
+info:
+  title: Event Notification using CloudEvents specifications
+  description: |
+    The event notification endpoint is used by the API server to notify the API consumer that an event occurred.  The notification is the message posted on listener side.
+
+    # Introduction
+
+    A lot of CAMARA APIs offer the capability to consumer to receive event.
+    Event data are defined in each API definition but in order to provide consistency across CAMARA APIs and to increase 
+    interoperability we will use [cloudevents](https://cloudevents.io/) specifications. In particular, CAMARA Event will 
+    be defined using [cloudevents-json-format](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/json-format.md)
+   
+    # Relevant terms and definitions
+
+      * **Occurrence** : An "occurrence" is the capture of a statement of fact during the operation of a software system.
+  
+      * **Event**: An "event" is a data record expressing an occurrence and its context. Events are routed from an 
+    event producer (the source) to interested event consumers.
+    
+      * **Producer**: The "producer" is a specific instance, process or device that creates the data structure 
+    describing the CloudEvent.
+  
+      * **Source**:  The "source" is the context in which the occurrence happened. In a distributed system it might 
+    consist of multiple producers. If a source is not aware of CloudEvents, an external producer creates 
+    the CloudEvent on behalf of the source.
+  
+      * **Consumer**: A "consumer" receives the event and acts upon it. It uses the context and data to execute some 
+    logic, which might lead to the occurrence of new events.
+  
+      * **Data**:  Domain-specific information about the occurrence (i.e. the payload). This might 
+    include information about the occurrence, details about the data that was changed, or more.
+  
+    
+    # API Functionality   
+      
+    Only one endpoint/operation is provided: `POST /cloudevents`
+    
+    This endpoint describes the event notification received on subscription listener side when the event occurred. A detailed description of the event notification is provided in the CAMARA API design guideline document.
+  
+  
+
+  termsOfService: http://swagger.io/terms/
+  contact:
+    email: project-email@sample.com
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.5.0-wip
+externalDocs:
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+security:
+  - oAuth2ClientCredentials: []
+
+servers:
+  - url: '{apiRoot}/{basePath}'
+    variables:
+      apiRoot:
+        default: http://localhost:8080
+        description: API root
+      basePath:
+        default: /event-notification/v0
+        description: Base path for the event notification API
+tags:
+  - name: Cloud Event
+    description: |
+      Events received on subscription listener side.
+paths:
+  /notifications:
+    post:
+      tags:
+        - CAMARA Cloud Event Notification
+      summary: "Notification endpoint to notify consumer that statement of fact had occurred"
+      description: Notification endpoint to notify consumer that statement of fact occurred. This endpoint must be exposed on the consumer side.
+      operationId: notifyCloudEvent
+      requestBody:
+        required: true
+        content:
+          application/cloudevents+json:
+            schema:
+              $ref: "#/components/schemas/CloudEvent"
+            examples:
+              QOS_STATUS_CHANGED_EXAMPLE:
+                $ref: '#/components/examples/QOS_STATUS_CHANGED_EXAMPLE'
+
+      responses:
+        201:
+          description: Created
+        202:
+          description: Accepted
+        400:
+          $ref: "#/components/responses/Generic400"
+        401:
+          $ref: "#/components/responses/Generic401"
+        403:
+          $ref: "#/components/responses/Generic403"
+        500:
+          description: Server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+        503:
+          $ref: "#/components/responses/Generic503"
+components:
+  securitySchemes:
+    oAuth2ClientCredentials:
+      type: oauth2
+      flows:
+        clientCredentials:
+          tokenUrl: '{tokenUrl}'
+          scopes: {}
+
+  schemas:
+    ErrorInfo:
+      type: object
+      required:
+        - status
+        - code
+        - message
+      properties:
+        status:
+          type: integer
+          description: HTTP response status code
+        code:
+          type: string
+          description: Code given to this error
+        message:
+          type: string
+          description: Detailed error description
+
+    DateTime:
+      type: string
+      format: date-time
+      description: Timestamp of when the occurrence happened. Must adhere to RFC 3339.
+      example: '2018-04-05T17:31:00Z'
+
+    Source:
+      type: string
+      format: uri-reference
+      minLength: 1
+      description: "Identifies the context in which an event happened(example: 'com.orange.camara.qod')"
+      example:
+        - https://github.com/cloudevents
+        - mailto:cncf-wg-serverless@lists.cncf.io
+        - urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
+        - cloudevents/spec/pull/123
+        - /sensors/tn-1234567/alerts
+        - 1-555-123-4567
+
+    CloudEvent:
+      description: The notification callback
+      required:
+        - id
+        - source
+        - specversion
+        - type
+      properties:
+        id:
+          type: string
+          description: identifier of this event, that must be unique in the source context.
+        source:
+          $ref: '#/components/schemas/Source'
+        type:
+          type: string
+          description: 'type of event as defined in each CAMARA API (examples: ROAMING_STATUS, QOS_STATUS_CHANGED, PAYMENT_COMPLETED, etc...)'
+          example: ROAMING_ON
+        specversion:
+          type: string
+          description: Version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version)
+        datacontenttype:
+          type: string
+          description: 'media-type that describes the event payload encoding, must be "application/json" for CAMARA APIs'
+        data:
+          type: object
+          description: Event details payload described in each CAMARA API and referenced by its type
+        subject:
+          description: Describes the subject of the event in the context of the event producer (identified by source).
+          type: string
+          example: ROAMING
+        time:
+          $ref: "#/components/schemas/DateTime"
+
+  responses:
+    Generic400:
+      description: Problem with the client request
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 400
+            code: "INVALID_ARGUMENT"
+            message: "Client specified an invalid argument, request body or query param"
+    Generic401:
+      description: Authentication problem with the client request
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 401
+            code: "UNAUTHENTICATED"
+            message: "Request not authenticated due to missing, invalid, or expired credentials"
+    Generic403:
+      description: Client does not have sufficient permission
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 403
+            code: "PERMISSION_DENIED"
+            message: "Client does not have sufficient permissions to perform this action"
+    Generic404:
+      description: Resource Not Found
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            code: NOT_FOUND
+            message: "The specified resource is not found"
+    Generic409:
+      description: Conflict
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            code: CONFLICT
+            message: "The specified resource is in a conflict"
+    Generic500:
+      description: Server error
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 500
+            code: "INTERNAL"
+            message: "Server error"
+    Generic503:
+      description: Service unavailable. Typically the server is down.
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 503
+            code: "UNAVAILABLE"
+            message: "Service unavailable"
+  examples:
+    QOS_STATUS_CHANGED_EXAMPLE:
+      value:
+        id: 123654
+        source: com.orange.camara.qod
+        type: QOS_STATUS_CHANGED
+        specversion: '1.0'
+        time: 2023-01-17T13:18:23.682Z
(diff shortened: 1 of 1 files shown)


patrice-conil:
Hi @rartych,
Thanks for your questions.
Thanks to @shilpa-padgaonkar for pointing this spec and @hdamker for insisting on reviewing it.
I probably missed some subtleties in the specs which are very rich, so don't hesitate to correct me.
I hope we will have many comments to enrich our OAS specification.

> Thank you @patrice-conil for this input, it is good starting point to evaluate Cloudevents (CE) for event notification format. From the quick look I have found some CE parameters that need to be discussed and decided about CAMARA defined content (based on [Cloudevents specification](https://github.com/cloudevents/spec/blob/v1.0/spec.md) ):
> 
> **source**
> 
>     * Type: `URI-reference`
> 
>     * An absolute URI is RECOMMENDED
>       _Should it reference general CAMARA Project, specific API or specific Provider Implementation?_
I think it should refer to the specific vendor implementation, i.e. the context in which the fact occurred
> 
> **type**
> 
>     * Type: String
> 
>     * SHOULD be prefixed with a reverse-DNS name. The prefixed domain dictates the organization which defines the semantics of this event type.
>       _Should it be something like org.camaraproject.api.qod or com.orange.camara.qod ?_
I think reverse DNS here is used to include package name with type to have fully qualified class name of the event type. Of course to have same reference for same type in all implementations we should use something like org.camara.api.qod.QosStatusChangedEvent 
> 
> **subject**
> 
>     * Type: String
> 
>     * OPTIONAL; If present, MUST be a non-empty string
> 
>     * Description: This describes the subject of the event in the context of the event producer (identified by source).
>       _Can we here specify the exact event type like QOS_STATUS_CHANGED ?_
> 
It's open to discussion.
> 
> Let's wait for evaluation and feedback from subprojects interested in Event notification.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,268 @@
...
+        description: API root
+      basePath:
+        default: /event-notification/v0
+        description: Base path for the event notification API
+tags:
+  - name: Cloud Event
+    description: |
+      Events received on subscription listener side.
+paths:
+  /notifications:
```
akoshunyadi, line 70: Shouldn't the name contain 'cloudevent'? Currently the name is general but the description is cloudevent-specific

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,268 @@
...
+        403:
+          $ref: "#/components/responses/Generic403"
+        500:
+          description: Server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorInfo"
+        503:
+          $ref: "#/components/responses/Generic503"
```
akoshunyadi, line 105: Some other response codes might be necessary to be [cloudevent-conform](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md#2-delivering-notifications)

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,268 @@
...
+        - type
+      properties:
+        id:
+          type: string
+          description: identifier of this event, that must be unique in the source context.
+        source:
+          $ref: '#/components/schemas/Source'
+        type:
+          type: string
+          description: 'type of event as defined in each CAMARA API (examples: ROAMING_STATUS, QOS_STATUS_CHANGED, PAYMENT_COMPLETED, etc...)'
```
akoshunyadi, line 167: I think we should use, as in request example, a fully qualified class name as type. The cloudevents spec also suggests to use a version.

artifacts/notification-as-cloud-event.yaml
patrice-conil: 200, 204, 410, 415 and 429 added

artifacts/notification-as-cloud-event.yaml
patrice-conil: done

akoshunyadi:
I think we will also need the property dataschema, otherwise how could the consumer process the data?

rartych:
[**dataschema**](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#dataschema) is optional attribute, so  it is not needed to send it with events.
Of course  the schema for the  `data` attribute needs to be defined - in API Design Guidelines or as a separate artifact to be available for both API and application developers, but not necessarily under specific URI.

patrice-conil:
Hi @rartych, @akoshunyadi,
If type contains fully qualified class name we can use it as discriminator and apply inheritance rules.

akoshunyadi:
@patrice-conil yes, but this spec doesn't contain any specific event details. And if we add a cloudevents subscription API that one is also generic. So we need a place for the definition of specific events, which we can do with inheritance from CloudEvent.

patrice-conil:
Hi @akoshunyadi, 
The idea is to provide a definition of EventDetails in each API (perhaps extend this specification in each API that provides Eventing?).
Perhaps we will also consolidate/aggregate all the definitions provided by the APIs into this specification.

rartych:
The CloudEvent example should be aligned with the final decisions on the content of each attribute in PR #56

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,313 @@
+openapi: 3.0.3
+info:
+  title: Event Notification using CloudEvents specifications
+  description: |
+    The event notification endpoint is used by the API server to notify the API consumer that an event occurred.  The notification is the message posted on listener side.
+
+    # Introduction
+
+    A lot of CAMARA APIs offer the capability to consumer to receive event.
```
PedroDiez, line 9:
```suggestion
    A lot of CAMARA APIs offer the capability to API consumer to receive events.
```

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,313 @@
...
+    Event data are defined in each API definition but in order to provide consistency across CAMARA APIs and to increase 
+    interoperability we will use [cloudevents](https://cloudevents.io/) specifications. In particular, CAMARA Event will
```
PedroDiez, line 11:
```suggestion
    interoperability we will use [cloudevents](https://cloudevents.io/) specifications. In particular, every CAMARA Event will 
```

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,313 @@
...
+      * **Consumer**: A "consumer" receives the event and acts upon it. It uses the context and data to execute some 
+    logic, which might lead to the occurrence of new events.
+  
+      * **Data**:  Domain-specific information about the occurrence (i.e. the payload). This might 
+    include information about the occurrence, details about the data that was changed, or more.
+  
+    
+    # API Functionality   
+      
+    Only one endpoint/operation is provided: `POST /cloudevents`
```
PedroDiez, line 37: In order to align with paths design, think we need to rename to:

`POST /your_webhook_notificationUrl`

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,313 @@
...
+  
+  
+
+  termsOfService: http://swagger.io/terms/
+  contact:
+    email: project-email@sample.com
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.5.0-wip
```
PedroDiez, line 49: When agreed (just doubt):
- Do we set initially to v0.5.0 (removing wip), doesn't?

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,313 @@
...
+    description: |
+      Events received on subscription listener side.
+paths:
+  /your_webhook_notificationUrl:
+    post:
+      tags:
+        - CAMARA Cloud Event notification endpoint
+      summary: "Cloud Event notification endpoint to notify consumer that statement of fact had occurred"
+      description: |
+        INFORMATIVE ENDPOINT: The value of this endpoint is freely declared by each client app by means of resource-based
```
PedroDiez, line 76: Just to provide further clarification and avoid misunderstanding. Suggest to add the following: @patrice-conil, @akoshunyadi, @rartych, @shilpa-padgaonkar 

"The value of this endpoint is freely declared by each client app by means of resource-based
        subscription or instance-based subscription. /your_webhook_notificationUrl is 
        just a convention naming referring to an absolute URL, indeed the one indicated by API client
        in the triggering of the procedure (resource-based or instance-based). _In this way, it represents an absolute URL, i.e.: notifications won't be sent to /event-notification/vX/your_webhook_notificationUrl_."

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,313 @@
...
+        - 1-555-123-4567
+
+    CloudEvent:
+      description: The notification callback
+      required:
+        - id
+        - source
+        - specversion
+        - type
+      properties:
```
PedroDiez, line 177: Open comment: 
- Should we align within CAMARA context naming format for source, type and subject fields? In "Cloudevent model", logically, is given the semantic for them, and in CAMARA maybe we need some alignment for an API client receive the same context of information

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,309 @@
...
+        description: API root
+      basePath:
+        default: /event-notification/v0
+        description: Base path for the event notification API
+tags:
+  - name: Cloud Event
+    description: |
+      Events received on subscription listener side.
+paths:
+  /cloudevents:
```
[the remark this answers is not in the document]
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
patrice-conil: Set to 0.1.0

artifacts/notification-as-cloud-event.yaml
patrice-conil: @PedroDiez, 
IMHO for the type and subject I think it is necessary that we all use the same one. For the context, it will probably have to be constructed in the same way but it will be different for each operator (one context per operator ?).

PedroDiez: @patrice-conil, do you mean use the same value in type and subject? (Maybe i am not getting this point)

PedroDiez: Thanks Patrice for the clarification

eric-murray:
Please provide any reviews or comments by 15th September

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+          type: string
+          description: 'type of event as defined in each CAMARA API (e.g. org.camara.api.qod.QosStatusChangedEvent)'
+          example: org.camara.api.qod.QosStatusChangedEvent
+        specversion:
+          type: string
+          description: Version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version)
+        datacontenttype:
+          type: string
+          description: 'media-type that describes the event payload encoding, must be "application/json" for CAMARA APIs'
+        data:
```
dfischer-tech, line 194: Shouldn't we also provide the "dataschema" attribute?
- Type: URI
- Description: Identifies the schema that data adheres to. Incompatible changes to the schema SHOULD be  reflected by a different URI. See [Versioning of Attributes in the Primer](https://github.com/cloudevents/spec/blob/v1.0/primer.md#versioning-of-attributes) for more information.
- Constraints:
    - OPTIONAL
    - If present, MUST be a non-empty URI

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+    CloudEvent:
+      description: The notification callback
+      required:
+        - id
+        - source
+        - specversion
+        - type
+      properties:
+        id:
+          type: string
```
dfischer-tech, line 180: as long the id MUST be a non-empty string - I also recommend the constraints "minLength" here.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+          description: identifier of this event, that must be unique in the source context.
+        source:
+          $ref: '#/components/schemas/Source'
+        type:
+          type: string
```
dfischer-tech, line 185: as long the type MUST be a non-empty string - I also recommend the constraints "minLength" here.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+          description: 'type of event as defined in each CAMARA API (e.g. org.camara.api.qod.QosStatusChangedEvent)'
+          example: org.camara.api.qod.QosStatusChangedEvent
+        specversion:
+          type: string
```
dfischer-tech, line 189: same here... "minLength" recommend

dfischer-tech (commented):
@patrice-conil - small review from my side Thx.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+      properties:
+        id:
+          type: string
+          description: identifier of this event, that must be unique in the source context.
+        source:
+          $ref: '#/components/schemas/Source'
+        type:
+          type: string
+          description: 'type of event as defined in each CAMARA API (e.g. org.camara.api.qod.QosStatusChangedEvent)'
+          example: org.camara.api.qod.QosStatusChangedEvent
```
maxl2287, line 187: I discovered that some Event-Handler-Platforms, which support Cloud-Events, do not support CamelCase and only allow lowercase: org.camara.api.qod.qosstatuschangedevent.

Just for your information

rartych: @maxl2287 Do you see the issue with more readable notation (with dashes) `org.camara.api.qod.qos-status-changed-event`?

akoshunyadi:
Currently the subject field seems be a duplication of the type. But actually the subject of the event in the QoD case is the session itself. So it could be "/sessions/6e8bc430-9c3a-11d9-9669-0800200c9a66" ?

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 503
+            code: "UNAVAILABLE"
+            message: "Service unavailable"
+  examples:
+    QOS_STATUS_CHANGED_EXAMPLE:
+      value:
+        id: 123654
```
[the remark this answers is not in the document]
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+      description: Client does not have sufficient permission
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 403
+            code: "PERMISSION_DENIED"
+            message: "Client does not have sufficient permissions to perform this action"
+    Generic404:
```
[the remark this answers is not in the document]
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+    Generic415:
+      description: Unsupported Media Type
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            code: UNSUPPORTED_MEDIA_TYPE
+            message: "The specified media type is not supported"
+    Generic429:
```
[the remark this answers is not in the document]
patrice-conil: 429 could be used... in case of client endpoint overload

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+      description: Too Many Requests
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            code: TOO_MANY_REQUESTS
+            message: "Endpoint does not support as many requests per time slot"
+    Generic500:
```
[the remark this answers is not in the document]
patrice-conil: @dfischer-tech  sure internal server will never happened ? 
I'll keep it just in case 😄 something goes wrong at server side.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+          $ref: "#/components/responses/Generic401"
+        403:
+          $ref: "#/components/responses/Generic403"
+        410:
+          $ref: "#/components/responses/Generic410"
+        415:
+          $ref: "#/components/responses/Generic415"
+        429:
+          $ref: "#/components/responses/Generic429"
+        500:
```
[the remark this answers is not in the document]
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
patrice-conil: As discussed in #56, 
type should be sufficient if used as discriminator.

artifacts/notification-as-cloud-event.yaml
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
patrice-conil: Done (min set to 15 the lenght of "org.camara.api." )

artifacts/notification-as-cloud-event.yaml
patrice-conil: Done, set to 3 (length of "1.0")

artifacts/notification-as-cloud-event.yaml
patrice-conil: Wait for #56 before using kebab-case or UPPER_SNAKE_CASE

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,314 @@
...
+    Generic404:
+      description: Resource Not Found
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            code: NOT_FOUND
+            message: "The specified resource is not found"
+    Generic409:
```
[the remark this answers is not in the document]
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,295 @@
...
+  - oAuth2ClientCredentials: []
+
+servers:
+  - url: '{apiRoot}/{basePath}'
+    variables:
+      apiRoot:
+        default: http://localhost:8080
+        description: API root
+      basePath:
+        default: /event-notification/v0
```
jlurien, line 63: As this is the basePath where the subscriptions client will receive the events, this may be whatever path that client registered as webhook.notificationUrl parameter in the subscriptions API, not necessarily  /event-notification/v0, right?

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,295 @@
...
+        - urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
+        - cloudevents/spec/pull/123
+        - /sensors/tn-1234567/alerts
+        - 1-555-123-4567
+
+    CloudEvent:
+      description: The notification callback
+      required:
+        - id
+        - source
```
jlurien, line 171: source is optional in CloudEvents

rartych: Source is in the "Required attributes" section of [spec](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#source-1)

artifacts/notification-as-cloud-event.yaml
patrice-conil: basePath removed and comment added in description

artifacts/notification-as-cloud-event.yaml
patrice-conil: api removed

artifacts/notification-as-cloud-event.yaml
jlurien: True, sorry, I was thinking in "subject".

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,291 @@
...
+        datacontenttype:
+          type: string
+          description: 'media-type that describes the event payload encoding, must be "application/json" for CAMARA APIs'
+        data:
+          type: object
+          description: Event details payload described in each CAMARA API and referenced by its type
+        subject:
+          description: Describes the subject of the event in the context of the event producer (identified by source).
+          type: string
+          example: ROAMING
```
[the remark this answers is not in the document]
rartych: Current DG  says that `subject` is "Not used in CAMARA notification."
It is optional attribute, so it will be supported by CloudEvents SDKs. The question is if it should be present in this OAS for notifications.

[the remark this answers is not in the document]
patrice-conil: Hi @akoshunyadi, @rartych,
So I'll remove it from this spec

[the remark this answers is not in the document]
jlurien: Agree to not make it normative by now. If we think of a usage for this property later we can add it

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,287 @@
...
+    email: project-email@sample.com
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.1.0-wip
+externalDocs:
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+security:
+  - oAuth2ClientCredentials: []
```
PedroDiez, line 54: Aligment would be required after resolution of [PR#57](https://github.com/camaraproject/Commonalities/pull/57/files)

artifacts/notification-as-cloud-event.yaml
PedroDiez: As the format agreed is org.camaraproject.<api-name>.<api-version>.
recommend to increase to "25" -> org.camaraproject.qod.v0.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,287 @@
...
+          type: string
+          description: Code given to this error
+        message:
+          type: string
+          description: Detailed error description
+
+    DateTime:
+      type: string
+      format: date-time
+      description: Timestamp of when the occurrence happened. Must adhere to RFC 3339.
```
PedroDiez, line 148: I would suggest to add in description something in this way (exact phrasing the one find more suitable), to highlight in CAMARA APIs this information is mandatory.

"Optional field inCloudEvents specification. Required in CAMARA APIs implementation"

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,287 @@
...
+  
+  
+
+  termsOfService: http://swagger.io/terms/
+  contact:
+    email: project-email@sample.com
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.1.0-wip
```
PedroDiez, line 49: Maybe we can remove "-wip", once we have ready for final review

PedroDiez (commented):
Some comments for alignment with guidelines

artifacts/notification-as-cloud-event.yaml
eric-murray: The API design guidelines allow the API consumer to have no security scheme (authorisation scheme) for event notifications. So no security also needs to be an option here.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,287 @@
...
+  description: Product documentation at CAMARA
+  url: https://github.com/camaraproject/
+security:
+  - oAuth2ClientCredentials: []
+
+servers:
+  - url: '{apiRoot}'
+    variables:
+      apiRoot:
+        default: http://localhost:8080
```
eric-murray, line 60: HTTPS is mandatory for event notifications, so would be good to reflect that in the default example

artifacts/notification-as-cloud-event.yaml
patrice-conil: Hi @PedroDiez, 
size increase to 25 and "WARN: This optional field in CloudEvents specification is required in CAMARA APIs implementation." added to DateTime

artifacts/notification-as-cloud-event.yaml
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
patrice-conil: Hi @eric-murray ,
I added : "Security is the responsibility of the API provider. The API provider must include its requirements in the security section of this specification or remove this section if it does not support any authentication scheme.
"
What do you think about?

eric-murray: @patrice-conil
I've seen no authentication scheme specified explicitly in a Swagger as `{}`
So I was expecting to see something like the following within the path definition:
```
      security:
        - oAuth2ClientCredentials: []
        - {}
```
In other words, either client credentials or no authorisation is allowed, but nothing else. When we need to send a notification, we would know which to use depending upon whether the client originally gave us a `notificationAuthToken` or not.

Anyway, this is just what I was expecting to see. If the client is expecting to receive notifications using some other authorisation scheme, they will be disappointed when they do not receive any.

patrice-conil: I missed this one! Done!

PedroDiez: Hi, 

@patrice-conil, @eric-murray regarding this thread. 

Two comments:
- scope: think we need to consider a scope for the use of this interface (e.g. notifications:create)

- security: think we should not specify security scheme within the interface and describe something like -> 
This interface can be used with securization (Authorization Header) or without it depending on the indication of `webhook.notificationAuthToken` in the involved procedure that generates associated event(s), applicable for both Instance-based or Resource-Based subscription models. So as when `webhook.notificationAuthToken` is not provided, no securization header will be indicated by menas of this interface.


Rationale: the way the API consumer generates the "notificationAuthToken" is on its discretion.

eric-murray: There is an open question that, if the API consumer provides an OAuth server for their notification endpoint and advertises that in their version of this YAML, should the API implementor use this to get a token, rather than mandate that the API consumer generate a token and provide this when they create the subscription?

If that is possible, then they will need to define the required scope (as well as providing the API implementor with a client id / secret). But I think the scope name is up to them. I don't see that CAMARA should impose their scope naming convention on the API consumer's own OAuth service.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,285 @@
...
+      description: |
+        Timestamp of when the occurrence happened. Must adhere to RFC 3339. 
+        WARN: This optional field in CloudEvents specification is required in CAMARA APIs implementation.
+      example: '2018-04-05T17:31:00Z'
+
+    Source:
+      type: string
+      format: uri-reference
+      minLength: 1
+      description: "Identifies the context in which an event happened(example: 'com.orange.camara.qod')"
```
rartych, line 153: URI-reference is not Reverse-DNS
```suggestion
      description: "Identifies the context in which an event happened - be a non-empty `URI-reference`.
```

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,285 @@
...
+      example:
+        - https://github.com/cloudevents
+        - mailto:cncf-wg-serverless@lists.cncf.io
+        - urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
+        - cloudevents/spec/pull/123
+        - /sensors/tn-1234567/alerts
+        - 1-555-123-4567
```
rartych, line 160: Here we have a list as the example, multiple examples are possible with `examples`
Or we can stay with one example like:
`      example: https://notificationSendServer12.supertelco.com`

```suggestion
     examples:
        absolute_uri_https:
          value: https://github.com/cloudevents
        absolute_uri_mailto:
          value: mailto:cncf-wg-serverless@lists.cncf.io
        URN_with_UUID:
          value: urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
        application_specific_relative_reference:
          value: /cloudevents/spec/pull/123
        application_specific:
          value: 1-555-123-4567
```

PedroDiez: fine with any of both approaches

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,285 @@
...
+        id:
+          type: string
+          description: identifier of this event, that must be unique in the source context.
+          minLength: 1
+        source:
+          $ref: '#/components/schemas/Source'
+        type:
+          type: string
+          description: 'type of event as defined in each CAMARA API (e.g.: org.camaraproject.qod.qos-status-changed-event)'
+          example: org.camaraproject.qod.qos-status-changed-event
```
PedroDiez, line 179:
```suggestion
          example: 'org.camaraproject.qod.qos-status-changed-event'
```

artifacts/notification-as-cloud-event.yaml
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,289 @@
...
+      description: |
+        Timestamp of when the occurrence happened. Must adhere to RFC 3339. 
+        WARN: This optional field in CloudEvents specification is required in CAMARA APIs implementation.
+      example: '2018-04-05T17:31:00Z'
+
+    Source:
+      type: string
+      format: uri-reference
+      minLength: 1
+      description: "Identifies the context in which an event happened - be a non-empty `URI-reference."
```
rartych, line 153: Closing backtick for `URI-reference`

```suggestion
      description: "Identifies the context in which an event happened - be a non-empty `URI-reference`."
```

patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
PedroDiez: @patrice-conil, @Eric regarding this point: Indicating `notificationsBearerAuth` scheme really shows how the notificationAuthToken is considered by Telco Operator in case of its reception in API request or explicit suscription.

With `oAuth2ClientCredentials: []`, is meant that the notificationAuthToken is an oAuth 2.0 client credentials token, and that is misleading because how that notificationAuthToken is generated by API consumer is not controlled neither decided by Telco Operator, it is up to the API consumer its generation.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,289 @@
...
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 503
+            code: "UNAVAILABLE"
+            message: "Service unavailable"
+  examples:
+    QOS_STATUS_CHANGED_EXAMPLE:
+      value:
+        id: "123e4567-e89b-12d3-a456-426655440000"
+        source: "com.orange.camara.qod"
```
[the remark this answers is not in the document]
patrice-conil: Done

artifacts/notification-as-cloud-event.yaml
patrice-conil: @PedroDiez, @eric-murray 
Suggestion from Pedro is integrated in last commit, Eric do you agree with that ?

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,287 @@
...
+        absolute_uri_https:
+          value: https://github.com/cloudevents
+        absolute_uri_mailto:
+          value: mailto:cncf-wg-serverless@lists.cncf.io
+        URN_with_UUID:
+          value: urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
+        application_specific_relative_reference:
+          value: /cloudevents/spec/pull/123
+        application_specific:
+          value: 1-555-123-4567
```
rartych, line 162: As in OAS prior to 3.1 multiple examples are not allowed in Schema Object I propose to extend the description with examples from CloudEvents specification and add simple example used in other places 
```suggestion
      description: |
        Identifies the context in which an event happened - be a non-empty `URI-reference` like:
        - URI with a DNS authority:
          * https://github.com/cloudevents
          * mailto:cncf-wg-serverless@lists.cncf.io
        - Universally-unique URN with a UUID:
          * urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66
        - Application-specific identifier:
          * /cloudevents/spec/pull/123
          * 1-555-123-4567
      example: "https://notificationSendServer12.supertelco.com"
```

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,286 @@
...
+          minLength: 3
+        datacontenttype:
+          type: string
+          description: 'media-type that describes the event payload encoding, must be "application/json" for CAMARA APIs'
+        data:
+          type: object
+          description: Event details payload described in each CAMARA API and referenced by its type
+        time:
+          $ref: "#/components/schemas/DateTime"
+
```
jlurien, line 194: To make the template even more useful, I would add also a sample schema for children events something like:

```
    ChildEventWithData:
      allOf:
        - $ref: "#/components/schemas/CloudEvent"
        - type: object
          properties:
            data:
              type: object
              description: Event details depending on the event type
              properties: {}
              required: {}
          required:
            - data
```

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,286 @@
...
+          example: 'org.camaraproject.qod.qos-status-changed-event'
+          minLength: 25
+        specversion:
+          type: string
+          description: Version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version)
+          minLength: 3
+        datacontenttype:
+          type: string
+          description: 'media-type that describes the event payload encoding, must be "application/json" for CAMARA APIs'
+        data:
```
jlurien, line 189: There is a discussion in https://github.com/camaraproject/QualityOnDemand/pull/224#discussion_r1387024248, about whether 1) to define `data` in the base class, as optional and with a generic `type: object`, or 2) to define `data` only in the child events which has data. Both would be correct, but not sure about which one would work better with code generators.

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,286 @@
...
+        INFORMATIVE ENDPOINT: The value of this endpoint is freely declared by each client app by means of resource-based
+        subscription or instance-based subscription. /your_webhook_notification_url is 
+        just a convention naming referring to an absolute URL, indeed the one indicated by API client
+        in the triggering of the procedure (resource-based or instance-based).  In this way, it represents an absolute
+        URL, i.e.: notifications won't be sent to /event-notification/vX/your_webhook_notification_url.
+      operationId: sendEvent
+      requestBody:
+        required: true
+        content:
+          application/cloudevents+json:
```
jlurien, line 79: I think this is correct but I think someone complained about lack of support by certain code generators. We could stick to just `application/json` if it is safer

rartych: According to [CloudEvents discussion](https://github.com/cloudevents/spec/issues/1118) format of "+json" is defined in RFC6839, so it should be supported by code generators.

 I propose to follow the CloudEvents [requirement](https://github.com/cloudevents/spec/blob/main/cloudevents/formats/json-format.md#3-envelope) for JSON format:

> Such a representation **MUST** use the media type `application/cloudevents+json`

artifacts/notification-as-cloud-event.yaml
```diff
@@ -0,0 +1,286 @@
...
+          * 1-555-123-4567
+      example: "https://notificationSendServer12.supertelco.com"
+
+    CloudEvent:
+      description: The notification callback
+      required:
+        - id
+        - source
+        - specversion
+        - type
```
rartych, line 169: In Design Guidelines we required the `time` atribute:
> Attribute `time` is tagged as optional in CloudEvents specification, but from CAMARA perspective we mandate to value this attributes.

```suggestion
        - type
        - time
```

artifacts/notification-as-cloud-event.yaml
rartych: I do not have experience with code generators, but currently we do not have any example of subproject using events without `data` object. So for simplicity I prefer approach 1) where we do not have to define child object  `ChildEventWithData` in each subproject using eventing, but `data` is in base class as optional.

jlurien: Indeed for that reason we have to define "a ChildEventWithData" schema, including `data` for each Child event. The name is just a templated one, real names we already have, among others:

QoD:
```
 EventQosStatusChanged:
      allOf:
        - $ref: "#/components/schemas/CloudEvent"
        - type: object
          properties:
            data: {}
```

DeviceStatus:
```
    EventRoamingStatus:
      description: event structure for roaming status change
      allOf:
        - $ref: "#/components/schemas/CloudEvent"
        - type: object
          properties:
            data:
              $ref: "#/components/schemas/RoamingStatus"

    EventRoamingOn:
      description: event structure for roaming on change
      allOf:
        - $ref: "#/components/schemas/CloudEvent"
        - type: object
          properties:
            data:
              $ref: "#/components/schemas/BasicDeviceEventData"
```

so we have to define the data per child event mandatorily always. The question was whether we should have to define it **also** in the base schema as optional.

patrice-conil: @jlurien, I agree with you, from an OO perspective we don't need to define it in the base schema.
I'm undecided because optional data field can also be considered a placeholder for what to extend in child classes.
But as you are the first that use this template in QoD and don't find it so useful, I will remove it from base class.
@rartych, both versions works with code generators.

jlurien: So far, it is kept as optional in the CloudEvent base class. I think this is what get more consensus. If both work well with code generators, they are totally equivalent and the final schema once the composition is applied is the same.

rartych: I am fine with both approaches. API Design Guidelines are "normative" for CAMARA, here we provide example that can be used by subprojects and API Consumers, but other definitions can be used if they conform to API Design Guidelines and CloudEvents specification.

rartych:
@jlurien @patrice-conil  Can we finalize still open points and go with final review, to be ready for Commonalities 0.2.0 release?

jlurien:
I see that you finally have removed  `data` from the base CloudEvent object. I thought you were going to finally keep it as optional, but it is OK to remove it  **as long as we add also an example about how to model a ChildEvent with required data**. Otherwise the artifact is not complete in order to serve as a model.

artifacts/notification-as-cloud-event.yaml
PedroDiez: Please see: https://github.com/camaraproject/Commonalities/pull/43#issuecomment-1830126520

patrice-conil:
Hi @rartych,
I think we haven't any open point (or I missed them) so we can finalize review.

rartych:
Thanks @patrice-conil .

@jlurien @PedroDiez @RubenBG7 @jordonezlucena Could you approve this PR, so we can be ready for release 0.2.0?

PedroDiez:
> Thanks @patrice-conil .
> 
> @jlurien @PedroDiez @RubenBG7 @jordonezlucena Could you approve this PR, so we can be ready for release 0.2.0?

will review today

PedroDiez (approved):
LGTM

RubenBG7 (approved):

jordonezlucena (approved):
/LGTM

PR #49, opened 2023-08-17
Add reserved word restrictions to API-design-guidelines.md

### LINKED ISSUE #29: Adding lists of code generator reserved words to the API Design Guidelines
**Problem description**
API code generators define lists of reserved words that cannot be used unquoted within various parts of an API definition without causing issues for the code generator. For example, the list of reserved words for python flask can be found [here](https://openapi-generator.tech/docs/generators/python-flask/#reserved-words). These reserved words can cause issues if used in the following parts of the API definition:
- Path and operation names
- Path/Query Parameter names
- Request and response body property names
- Security schemes
- Component names
- OperationID

If an API does include reserved words in those parts of the API definition, then the API implementor will have to modify the API definition before feeding it to the code generator. For some uses of reserved words (e.g. path, parameter or property names) such changes cannot be made without changing the API definition itself.

**Expected action**
A list of reserved words should be maintained within the CAMARA [API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md) along with rules as to where these reserved words cannot be used. The reserved words themselves can either be explicitly copied into the design guidelines, or linked to the code generator tool's own specification. Users of code generators other than python flask would need to raise their own PRs to add the relevant list of reserved words.

**Additional context**
API definitions are already being modified to make life simpler for code generators. This is a small additional issue with code generators, and one easily avoided. Currently I am not aware of any CAMARA API which does use reserved words in a way that would trip up python flask, so it would seem unlikely that changes to current API definitions would be required as a result of this change.

eric-murray:
#### What type of PR is this?
* documentation

Adds restrictions on use of Open API generator reserved words in an API specification, and adds a link to the current reserved word list for Python Flask.

Fixes #29

Additional reserved word list can be added by other developers

#### Changelog input

```
 release-note
- Adds restrictions on use of Open API generator reserved words in an API specification
- Adds a link to the current reserved word list for Python Flask.
```

#### Additional documentation 
```
docs
- [Python Flask Reserved Words](https://openapi-generator.tech/docs/generators/python-flask/#reserved-words)
```

### DIFF
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -966,7 +966,7 @@ Below considerations should be checked when an API is documented:
     - [OpenAPI Generator (Java)](https://openapi-generator.tech/docs/generators/java/#reserved-words)
     - [OpenAPI Generator (Go)](https://openapi-generator.tech/docs/generators/go/#reserved-words)
     - [OpenAPI Generator (Kotlin)](https://openapi-generator.tech/docs/generators/kotlin/#reserved-words)
-    - [OpenAPI Generator (Swift)](https://openapi-generator.tech/docs/generators/swift5#reserved-words)
+    - [OpenAPI Generator (Swift5)](https://openapi-generator.tech/docs/generators/swift5#reserved-words)
 
 ### 11.1 General Information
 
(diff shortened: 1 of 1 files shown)


patrice-conil (changes requested):
I think we could also add reserved keywords for JAVA and GO for consuming or exposing APIs from backends.
https://openapi-generator.tech/docs/generators/java/#reserved-words
https://openapi-generator.tech/docs/generators/go/#reserved-words

But also kotlin and swift for API consumption from mobiles
https://openapi-generator.tech/docs/generators/kotlin/#reserved-words
https://openapi-generator.tech/docs/generators/swift5#reserved-words

patrice-conil (approved):
/LGTM

PR #50, opened 2023-08-17
Remove option to use HTTP HEAD from API-design-guidelines.md

### LINKED ISSUE #28: Proposal to remove the option of HTTP HEAD requests from the API Design Guidelines
**Problem description**
The current [API Design Guidelines](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md) allow for HTTP HEAD requests to be defined for CAMARA APIs. Currently no CAMARA API defines these, and for many APIs it would be anyway be problematic to implement these as the API is not returning a static resource. Response bodies of CAMARA APIs are typically small, so there is no great advantage in knowing the response size in advance.

Hence it is concluded that there is no good use case for allowing HEAD requests in CAMARA APIs, but there remains the risk it could be included in future by an API because it "seemed like a good idea", creating a tricky problem for API implementors.

**Expected action**
Use of HTTP HEAD requests should be explicitly prohibited by the CAMARA API Design Guidelines.

**Additional context**
None

eric-murray:
#### What type of PR is this?

* documentation

HEAD requests are not suitable for network APIs managing dynamic data and there use should be prohibited. There are alternatives to HEAD requests if the purpose is to see whether the data has changed since last requested - for example, the `If-Modified-Since` header.

Fixes #28

No APIs currently define HEAD requests

#### Changelog input
```
 release-note
 - Remove option to use HTTP HEAD from API design guidelines
```

#### Additional documentation 
N/A

### DIFF
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -258,7 +258,6 @@ PUT |  Update | Replaces a specific resource. Returns the resource URL when the 
 DELETE | Delete | Delete a specific resource.  |
 PATCH |  Update | Updates a specific resource, applying all changes at the same time. If resource does not exist, it will be created. Returns the resource URL when the update ends. If any error occurs during the update, all of them will be cancelled.   |
 OPTIONS | Read | Returns a 200 OK with an allowed methods list in the specific resource destined to the header allowed joined to an HTML document about the resource + an API Doc link.  |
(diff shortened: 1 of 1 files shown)


patrice-conil (approved):
LGTM

rartych (approved):

Issue #44, opened 2023-07-28
API Design for "Resource-based subscription"


PedroDiez:
**Problem description**
Within Commonalities API design guidelines, it has been indicated a common way to define the support for notifications model. 
https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#12-subscription-notification--event

In "Resource-based subscription" when applying this model to the API design, several approaches are seen:
Model 1:
- Device Location Verification (separate API): https://github.com/camaraproject/DeviceLocation/pull/74/files

Model 2:
- Device Status (functionlity embebbed within API): https://github.com/camaraproject/DeviceStatus/blob/main/code/API_definitions/device-status.yaml#L134

We think it is reasonable to have a common way to design this features.
Our proposal is to align the approach and use a common model when including this features in WG APIs families.

This will require some indication in Commonalities Guidelines. Once some conclusion is reached.

**Possible evolution**
Our proposal is to follow Model 1 when designing "Resource-based subscription" in API families. Main reasons to support this model are:

- Design 'simpler' APIs, avoiding APIs with many endpoints, that make its implementation and availability complex.

- Have more 'flexibility' to evolve "Resource-based subscription" separately, so also roadmaps can be decoupled in order to expose this functionality. In this way, timeline for API functionality and subscriptions feature can be decoupled so Telco Operators can provide them in a more agile way depending on its market needs.

- Notification subscriptions systems can be from technical perspective decoupled from API BE Logic, so having different "server url" exposition. That is, the logic of subscription management is managed separately from the API BE logic, as the main "core" of this logic is to identify when an event occurs if there is a matching susbcription for that event and if so, send a notification to the notificationURL indicated   

Would like to see the view about this point  within CAMARA

PedroDiez:
@shilpa-padgaonkar, @hdamker, @eric-murray, @patrice-conil, @bigludo7
RubenBG7

cc @RubenBG7, @jpengar, @jlurien

patrice-conil:
HI @PedroDiez,
We (@bigludo7 and I) also believe that model 1 is the way to go to clearly define that the /notifications endpoint will be called from the consumer side of the API.
As this model is not well supported by code generators (openapi-generator or swagger-generator for example), we proposed to define a separate OAS specification for the /notifications endpoint that consumers could use to generate their server part.
We have made two joint proposals in commonalities that you can review and comment on.

PedroDiez:
Thanks @patrice-conil, I will review them

[...]

Issue #10, opened 2023-06-28
Attribute for filtering in GET request


bigludo7:
One small issue but we have the point in Carrier billing API with @PedroDiez.

We have to provide filtering in GET request for Payment (for example; to request a API consumer to retrieve all its Payments done from June 1st to June 8th and getting in the response only payment id & state)

The API design guideline describes this in paragraph 7.2  but did not mention explicitly the query parameter **attribute name** that we must define in the swagger. For consistency we should use same name in all API for this attribute.

Do we want to use `fields `like in TMF ? or other name ? 

Welcome to any any/suggestion

PedroDiez:
Regarding this point, as per my understanding, the meaning of filtering is to restrict the range of parameter values that could be returned, in the same fashion as commented by Eric in his first comment.

An about the pattern, I believe pattern 2 is more flexible for each API needs, due to each API can have its particularities (in some APIs can make sense to filter by a range of dates, in other APIs to filter based on some values that a given param can take in response -e.g. enum values-,...). In this way, each initiative can work on defining the filter criteria needed and the suitable naming for that purpose

patrice-conil:
Another option could be to expose some resources using [GrapQL](https://graphql.org/learn/) to be able to exactly match consumer expectations. What is quite easy when you use Quarkus or Spring frameworks.
I think it will solve the first point (deep field filtering, filtering by fixed value) but not the second (filtering by ranges or inequality)

rartych:
According to this [answer from Swagger](https://community.smartbear.com/t5/SwaggerHub-Questions/Swagger-compliance-and-queryParameter-suffix-filtering-example/m-p/219422/highlight/true#M1454) pattern 2 is possible but not straightforward solution.

It also suggest another reason for OAS 3.1 https://github.com/camaraproject/Commonalities/issues/11

PR #56, opened 2023-09-01
Update DG with use of callbacks & cloudEvents

### LINKED ISSUE #8: Provides specific CAMARA OAS for notification event

bigludo7:
#### What type of PR is this?

* enhancement/feature

In this PR several notifications update are provided:

- enforce use of `callbacks` in CAMARA API to describe notification event structure
- use of **CloudEvents** format for CAMARA API

Additionally this PR should be accompanied with PR #43.
This PR triggered cancellation of PR #41  

Fixes #8 #44 
Note: Issue 44 covers several topics. One topic is about general ruling to have separate (or not) swagger for resource-based subscription from the nominal swagger for the API. This point is not tackled in this PR and must be followed in a new issue.

#### Changelog input

```
- enforce use of `callbacks` in CAMARA API to describe notification event structure
- use of CloudEvents format for CAMARA API

```

#### Additional documentation 

This section can be blank.

```
docs

```


documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1288,51 @@ response:
...
+Note: If the API provides both patterns (indirect and resource-based), and the API customer requests both (instance based + subscription), the 2 requests should be handled independently & autonomously. Depending on server implementation, it is acceptable, when the event occurs, that one or two notifications are sent to listener.
 
 
 ### 12.2 Event notification
 
 The event notification endpoint is used by the API server to notify the API consumer that an event occured.
 
-Note: The notification is the message posted on listener side. We describe the notification message in the CAMARA OAS but it could confusing because this endpoint should be implemented on the business API consumer side. This notice should be explicited mentioned in all CAMARA API documentation featuring notifications.
+CAMARA event notification leverages **CloudEvents** as it is a vendor-neutral specification for defining the format of event data. A generic neutral CloudEvent notification swagger is available in commonalities/artifact directory (notification-as-cloud-event.yaml).
```
rartych, line 1306: I propose to add reference to CloudEvents main webpage (or specification)

```suggestion
CAMARA event notification leverages **[CloudEvents](https://cloudevents.io/)** as it is a vendor-neutral specification for defining the format of event data. A generic neutral CloudEvent notification swagger is available in commonalities/artifact directory (notification-as-cloud-event.yaml).
```

documentation/API-design-guidelines.md
```diff
@@ -1353,20 +1361,22 @@ curl -X 'POST' \
...
-      "uePort": 5060,
-      "roaming": true,
-      "countryCode": 208,
-      "countryName": "FR"
-    }
-  }
+  "id": 123654,
+  "source": "supertelco.notificationSendServer12",
+  "type": "org.camara.api.device-status.RoamingStatusEvent",
+  "specversion": 1,
```
rartych, line 1367: String type

```suggestion
  "specversion": "1.0",
```

documentation/API-design-guidelines.md
```diff
@@ -1389,18 +1399,20 @@ curl -X 'POST' \
...
-        "ipv4Addr": "192.168.0.1"
-      },
-      "uePort": 5060,
-      "terminationReason": "Service terminates for lack of consent"
-    }
-  }
+  "id": 123658,
+  "source": "supertelco.notificationSendServer12",
+  "type": "org.camara.api.device-status.RoamingSubscriptionEndsEvent",
+  "specversion": 1,
```
rartych, line 1405: String type

```suggestion
  "specversion": "1.0",
```

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1288,51 @@ response:
...
-| eventDetail | object | Event details structure depending on the eventType | mandatory |
+| eventSubscriptionId | string | The event subscription identifier - must be valued for Resource-based subscription | optional |
+
 
 Note: For operational and troubleshooting purposes it is relevant to accommodate use of `X-Correlator` header attribute. API listener implementations have to be ready to support and receive this data.
 
-Specific eventType "SUBSCRIPTION_ENDS" is defined to inform listener about subscrition termination. It is used when the subscription expire time (required by the requester) has been reached or if the API server has to stop sending notification prematurely. For this specific event, the `eventDetail` must feature `terminationReason` attribute.
+Specific event notification subject "EventSubscriptionEnds" is defined to inform listener about subscrition termination. It is used when the subscription expire time (required by the requester) has been reached or if the API server has to stop sending notification prematurely. For this specific event, the `data` must feature `terminationReason` attribute.
```
PedroDiez, line 1335: Would keep "UPPER_SNAKE_CASE"

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1288,51 @@ response:
...
+| type | string | a value describing the type of event related to the originating occurrence. For consistency accross API we mandate following pattern: org.camara.api._name of the API_._event name_ (for exemple org.camara.api.device-status.RoamingStatusEvent ) | mandatory |
+| specversion | string | version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version) | mandatory |
+| datacontenttype | string | media-type that describes the event payload encoding, must be `application/json for CAMARA APIs` | optional |
+| subject | string | describes the subject of the event. In CAMARA we enforce to use subject as defined in each API. This attribute is tagged as optional in CloudEvents specification but from CAMARA perspective we **strongly** recommend to value this attribute. | optional |
+| time | string  datetime| timestamp of when the occurrence happened (must adhere on CAMARA datetime recommendation based on RFC 3339) | optional |
```
PedroDiez, line 1323: Think that in CAMARA we have to consider "time" mandatory, as we are using this replacing "former" model within SusbcriptionDetail

documentation/API-design-guidelines.md
```diff
@@ -1215,7 +1216,7 @@ The `subscriptionDetail` must have at least an eventType attribute:
 
 | name | type | attribute description | cardinality |
 | ----- |	-----  |	 -----  |  -----  | 
-| eventType | string | Type of event subscribed. This attribute must be present in the `POST` request. It is open to API working group to allow providing a list of event type based on specific UC. `eventType` must follow UPPER_SNAKE_CASE format. | mandatory  |
+| subject | string | Type of event subscribed. This attribute must be present in the `POST` request. It is open to API working group to allow providing a list of event subject based on specific UC. `subject` must follow CamelCase format. | mandatory  |
```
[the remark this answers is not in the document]
bigludo7: @PedroDiez 

Thanks for the review. As a lot of your comments are about `type` and `subject` usage I tried here to explain our thinking.

Using `type` attribute was our initial approach (with @patrice-conil) but crafting the device status API we got a problem with this approach and proposed to switch to `subject` (to make the thing simpler for implementation & consumers).

Let me explain why:

- Accordingly to cloudEvents [specification ](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md) the `type` contains a value describing the type of event related to the originating occurrence and should be prefixed with a reverse-DNS name. So the value should be something like this `org.camara.api.device-status.RoamingStatusEvent`. 
- For a lot of API we will have several types of event. In the swagger of the API we need to provide for each event type a data structure. This data structure extend event structure with specific attribute related to the event type. In order to do that we use extension pattern with discriminator 
-  With the `type` value it seems to us too complex to use this `type` field as a `discriminator` - that's why we proposed to use `subject` to store the event type.  This is valid approach as cloudEvents specification define subject as "_the subject of the event in the context of the event producer_" 
- Last but not least as this _subject_ is use a discriminator it made sense for us to use same format (CamelCase) as the one used in the swagger for the classes.  

We crafted the device status API with this new pattern and the CloudEvent class definition will be:

```yaml
    CloudEvent:
      description: The notification callback
      required:
        - id
        - source
        - specversion
        - type
      properties:
        id:
          type: string
          description: identifier of this event, that must be unique in the source context.
        source:
          $ref: '#/components/schemas/Source'
        type:
          type: string
          description: 'type of event as defined in each CAMARA API'
          example: org.camara.api.device-status.RoamingStatusEvent
        specversion:
          type: string
          description: Version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version)
          example: 1.0
        datacontenttype:
          type: string
          description: 'media-type that describes the event payload encoding, must be "application/json" for CAMARA APIs'
          example: application/json
        data:
          type: object
          description: Event details payload described in each CAMARA API and referenced by its type
        subject:
          description: Describes the subject of the event in the context of the event producer (identified by source).
          type: string
        time:
          $ref: "#/components/schemas/DateTime"
      discriminator:
        propertyName: 'subject'
        mapping:
            EventRoamingStatus: '#/components/schemas/EventRoamingStatus'
            EventRoamingChangeCountry : '#/components/schemas/EventRoamingChangeCountry'
            EventRoamingOff: '#/components/schemas/EventRoamingOff'
            EventRoamingOn: '#/components/schemas/EventRoamingOn'
            EventSubscriptionEnds : '#/components/schemas/EventSubscriptionEnds'`
```

If we use the 'type'  it will be something like this for the extension which seems to us not a good swagger design:

``` yaml
      discriminator:
        propertyName: 'subject'
        mapping:
            org.camara.api.device-status.RoamingStatusEvent: '#/components/schemas/EventRoamingStatus'
            org.camara.api.device-status.RoamingChangeCountryEvent : '#/components/schemas/RoamingChangeCountry'
            org.camara.api.device-status.RoamingOffEvent: '#/components/schemas/EventRoamingOff'
            org.camara.api.device-status.RoamingOnEvent: '#/components/schemas/EventRoamingOn'
            org.camara.api.device-status.SubscriptionEndsEvent : '#/components/schemas/EventSubscriptionEnds
```

WDYT?

documentation/API-design-guidelines.md
bigludo7: Do you think we can be more restrictive than CloudEvnts ? 
We can add in the description "_This attribute is tagged as optional in CloudEvents specification but from CAMARA perspective we **strongly** recommend to value this attribute_."

rartych: I think we can define "CAMARA profile" of CloudEvents - the change from _optional_ to _mandatory_ or _recommended_ to _required_ does not complicate development of API implementations or applications, it can even simplify the task.

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1288,51 @@ response:
...
 
-For consistence among CAMARA APIs the uniform `eventNotifications` model must be used:
+For consistence among CAMARA APIs the uniform CloudEvents model must be used with following rules:
 
 | name | type | attribute description | cardinality |
 | ----- |	-----  |	 -----  |  -----  | 
-| eventSubscriptionId | string | subscription identifier - could be valued for Resource-based subscription | optional |
-| event | object | event structure - see next table | mandatory |
+| id | string | identifier of this event, that must be unique in the source context. | mandatory |
+| source | string - URI | identifies the context in which an event happened in the specific Provider Implementation. Often this will include information such as the type of the event source, the organization publishing the event or the process that produced the event. The exact syntax and semantics behind the data encoded in the URI is defined by the event producer. | mandatory |
```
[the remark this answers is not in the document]
PedroDiez: Please discard my previous comment regarding the format, as it would not be compliant to CloudEvents specs.

As commented by Rafał, the recommendation is an absolute URI. If we are ok to go for this recommendation at least it needs to be indicated in the guidelines for any Operator to follow that format.

documentation/API-design-guidelines.md
PedroDiez: Get your points Ludovic, Rafał. 

I think at least we can add a sentence in that way, just to reflect the relevance of this info. 
If quorom, also imho could be mandatory (not a real problem being optional). Some rationale below:

In TEF:
- From technical perspective we can implement it
- From Business perspective, we think is really valuable information as it indicates when the event occurs. There can be problems in systems, retries,...and this information guarantees API consumer when the event really happened, because it may not consider that the reception timestamp is the event timestamp and in some business cases such information can be the "key" one (e.g. fraud prevention, enter into roaming,...)

bigludo7 (commented):
Reworked `data` attribute description following @PedroDiez comment

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1288,51 @@ response:
...
+| specversion | string | version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version) | mandatory |
+| datacontenttype | string | media-type that describes the event payload encoding, must be `application/json for CAMARA APIs` | optional |
+| subject | string | describes the subject of the event. In CAMARA we enforce to use subject as defined in each API. This attribute is tagged as optional in CloudEvents specification but from CAMARA perspective we **strongly** recommend to value this attribute. | optional |
+| time | string  datetime| timestamp of when the occurrence happened (must adhere on CAMARA datetime recommendation based on RFC 3339) | optional |
+| data | object| event notification details payload described in each CAMARA API and referenced by its `subject` | optional |
 
-Following table defines event attribute object structure: 
+`data` structure is dependant to each API but we mandate presence of this following attribute for all event notification
```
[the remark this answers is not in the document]
bigludo7: Agree - updated.

documentation/API-design-guidelines.md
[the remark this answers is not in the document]
PedroDiez: Understood the point of using "subject" as a discriminator in order to model for both the subscriptions and the events. 
In the second approach propertyName would be "type".

As "type" has reverse-DNS format, it is "long" name and it is a matter of Look&Feel.

And the constraint of doing that is that "subject" needs to be settled as mandatory in our model, because a field being used as discriminator  has to be mandatory (now i understand that this model allows in the case of the subscription to indicate the related information based on eventType, I thought this only refers to event_type and additional info needed for subscription).

So basically, we could do this using "type" without the need to use this additional field.


Regarding "type" format, the comment about using UPPER_SNAKE_CASE was in the line of doing something like that:

org.camara.api.device-status.ROAMING_STATUS (so the eventType is clearly identified in the context)

        mapping:
            org.camara.api.device-status.ROAMING_STATUS: '#/components/schemas/EventRoamingStatus'

bigludo7 (commented):
Changed `time ` &  `subject `cardinality in Notification event schema following @rartych & @PedroDiez guidance.

documentation/API-design-guidelines.md
bigludo7: updated accordingly to you comments (switched both to mandatory)

eric-murray:
Please provide any additional reviews or comments by 15th September

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1288,51 @@ response:
...
+For consistence among CAMARA APIs the uniform CloudEvents model must be used with following rules:
 
 | name | type | attribute description | cardinality |
 | ----- |	-----  |	 -----  |  -----  | 
-| eventSubscriptionId | string | subscription identifier - could be valued for Resource-based subscription | optional |
-| event | object | event structure - see next table | mandatory |
+| id | string | identifier of this event, that must be unique in the source context. | mandatory |
+| source | string - URI | identifies the context in which an event happened in the specific Provider Implementation. Often this will include information such as the type of the event source, the organization publishing the event or the process that produced the event. The exact syntax and semantics behind the data encoded in the URI is defined by the event producer. | mandatory |
+| type | string | a value describing the type of event related to the originating occurrence. For consistency accross API we mandate following pattern: org.camara.api._name of the API_._event name_ (for exemple org.camara.api.device-status.RoamingStatusEvent ) | mandatory |
```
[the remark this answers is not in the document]
rartych: Taking into account this https://github.com/camaraproject/Commonalities/pull/43#discussion_r1319631841 and the convention for DNS and reverse-DNS names to use lowercase :
```suggestion
| type | string | a value describing the type of event related to the originating occurrence. For consistency accross CAMARA APIs we mandate following pattern of reverse-DNS written in lowercase: org.camara.api._name of the API_._event name_ (for exemple org.camara.api.device-status.roamingstatusevent ) | mandatory |
```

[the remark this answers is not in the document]
rartych: The option is to use more readable kebab-case i.e.  `org.camara.api.device-status.roaming-status-event`

documentation/API-design-guidelines.md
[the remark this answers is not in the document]
jlurien: `type` is clearly intended to be used in CloudEvents as the identifier of the type of event everywhere, and it determines the content of data, so it is the right choice for discriminator. I think this would not be discussed if type didn't require to be expressed in reverse-DNS format, which is not the best name for a schema. But discriminator's mapping comes to solve this, so we can choose whatever schema name is right, without needing to distort the CloudEvents spec.

`subject` is optional precisely because is not needed in every situation. It is not even required to have the same value for all events of the same type. We may discuss if it is needed in CAMARA, but it should not be used as discriminator.

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1288,54 @@ response:
...
+| specversion | string | version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version) | mandatory |
+| datacontenttype | string | media-type that describes the event payload encoding, must be `application/json for CAMARA APIs` | optional |
```
rartych, line 1321: `datacontenttype="application/json"` is default for CloudEvents, I guess it is a typo here
```suggestion
| datacontenttype | string | media-type that describes the event payload encoding, must be `application/json` for CAMARA APIs| optional |
```

documentation/API-design-guidelines.md
```diff
@@ -1353,20 +1364,22 @@ curl -X 'POST' \
...
-      },
-      "uePort": 5060,
-      "roaming": true,
-      "countryCode": 208,
-      "countryName": "FR"
-    }
-  }
+  "id": 123654,
+  "source": "supertelco.notificationSendServer12",
+  "type": "org.camara.api.device-status.RoamingStatusEvent",
```
rartych, line 1369:
```suggestion
  "type": "org.camara.api.device-status.roamingstatusevent",
```

documentation/API-design-guidelines.md
```diff
@@ -1389,18 +1402,20 @@ curl -X 'POST' \
...
-      "ueId": {
-        "ipv4Addr": "192.168.0.1"
-      },
-      "uePort": 5060,
-      "terminationReason": "Service terminates for lack of consent"
-    }
-  }
+  "id": 123658,
+  "source": "supertelco.notificationSendServer12",
+  "type": "org.camara.api.device-status.RoamingSubscriptionEndsEvent",
```
rartych, line 1407:
```suggestion
  "type": "org.camara.api.device-status.roamingsubscriptionendsevent",
```

documentation/API-design-guidelines.md
```diff
@@ -1389,18 +1402,20 @@ curl -X 'POST' \
...
-    "eventDetail": {
-      "ueId": {
-        "ipv4Addr": "192.168.0.1"
-      },
-      "uePort": 5060,
-      "terminationReason": "Service terminates for lack of consent"
-    }
-  }
+  "id": 123658,
+  "source": "supertelco.notificationSendServer12",
```
rartych, line 1406: `source` MUST be a non-empty URI-reference
```suggestion
  "source": "https://notificationSendServer12.supertelco.com",
```

documentation/API-design-guidelines.md
```diff
@@ -1353,20 +1364,22 @@ curl -X 'POST' \
...
-        "ipv4Addr": "192.168.0.1"
-      },
-      "uePort": 5060,
-      "roaming": true,
-      "countryCode": 208,
-      "countryName": "FR"
-    }
-  }
+  "id": 123654,
+  "source": "supertelco.notificationSendServer12",
```
rartych, line 1368: `source` MUST be a non-empty URI-reference
" The application MAY use UUIDs, URNs, DNS authorities or an application-specific scheme to create unique `source `identifiers."

```suggestion
"source": "https://notificationSendServer12.supertelco.com",
```

rartych:
Despite my updates in suggestions above I wonder if current model of `type `and `subject`is redundant.
Would it be sufficient to have for example:
```
"type": "org.camara.api.device-status"
"subject": "EventRoamingStatus"
```
CloudEvents considers [versioning](https://github.com/cloudevents/spec/blob/v1.0/primer.md#versioning-of-attributes) in `type` then we can have:
```
"type": "org.camara.api.device-status.v1"
"subject": "EventRoamingStatus"
```
It may be useful, but can complicate the processing and is rather limited to major version number.

documentation/API-design-guidelines.md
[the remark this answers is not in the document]
bigludo7: Thanks @jlurien for the effort in checking CloudEvents documentation.
Let's move in the direction you suggest.

I will change accordingly the guideline by using `type` as discriminator and not `subject`.

About `subject` probably we could keep this one for later and not define any usage as of now in the guideline.
cc: @rartych

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1288,51 @@ response:
...
+| type | string | a value describing the type of event related to the originating occurrence. For consistency accross API we mandate following pattern: org.camara.api._name of the API_._event name_ (for exemple org.camara.api.device-status.RoamingStatusEvent ) | mandatory |
+| specversion | string | version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version) | mandatory |
+| datacontenttype | string | media-type that describes the event payload encoding, must be `application/json for CAMARA APIs` | optional |
+| subject | string | describes the subject of the event. In CAMARA we enforce to use subject as defined in each API. This attribute is tagged as optional in CloudEvents specification but from CAMARA perspective we **strongly** recommend to value this attribute. | optional |
+| time | string  datetime| timestamp of when the occurrence happened (must adhere on CAMARA datetime recommendation based on RFC 3339) | optional |
+| data | object| event notification details payload described in each CAMARA API and referenced by its `subject` | optional |
```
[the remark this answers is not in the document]
bigludo7: As of now you're probably right but I prefer to keep this flexible here in the guideline and let Working Group enforce data presence (and attributes)  to each event type.

bigludo7:
@jlurien @rartych Thanks a lot for the help & review.
I hope to have accommodate the guideline following your recommendations.
May I ask to take another look to check if we're aligned or at least check we do not have any blocker on your side.

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1290,54 @@ response:
...
+For consistence among CAMARA APIs the uniform CloudEvents model must be used with following rules:
 
 | name | type | attribute description | cardinality |
 | ----- |	-----  |	 -----  |  -----  | 
-| eventSubscriptionId | string | subscription identifier - could be valued for Resource-based subscription | optional |
-| event | object | event structure - see next table | mandatory |
+| id | string | identifier of this event, that must be unique in the source context. | mandatory |
+| source | string - URI | identifies the context in which an event happened in the specific Provider Implementation. Often this will include information such as the type of the event source, the organization publishing the event or the process that produced the event. The exact syntax and semantics behind the data encoded in the URI is defined by the event producer. | mandatory |
+| type | string | a value describing the type of event related to the originating occurrence. For consistency accross API we mandate following pattern: `org.camaraproject.<api-name>.<EVENT_NAME` (for exemple org.camaraproject.device-status.ROAMING_CHANGE_COUNTRY ) | mandatory |
```
shilpa-padgaonkar, line 1321: As this attribute contains a value describing the type of event related to the originating occurrence. The type of the event could be prefixed before the API name
org.camaraproject.<type/category>.<api-name>.<EVENT_NAME>

for eg.
org.camaraproject.network.qod.<EVENT_NAME>
org.camaraproject.device.device-status.<EVENT_NAME>
org.camaraproject.device.verify-location.<EVENT_NAME>

This could later align well our agenda number 4? WDYT?

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1290,54 @@ response:
...
+CAMARA event notification leverages **[CloudEvents](https://cloudevents.io/)** as it is a vendor-neutral specification for defining the format of event data. A generic neutral CloudEvent notification swagger is available in commonalities/artifact directory (notification-as-cloud-event.yaml).
+
+Note: The notification is the message posted on listener side. We describe the notification(s) in the CAMARA API using the `callbacks`. From API consumer it could confusing because this endpoint must be implemented on the business API consumer side. This notice should be explicited mentioned in all CAMARA API documentation featuring notifications.
 
-Only Operation POST is provided for `eventNotifications` and the expected response code is `204`.
+Only Operation POST is provided for `eventNotifications` and the expected response code is `204`. 
+The url for this `POST` operation must be specified in the swagger as `{$request.body#/webhook/notificationUrl}`. 
 
-For consistence among CAMARA APIs the uniform `eventNotifications` model must be used:
+For consistence among CAMARA APIs the uniform CloudEvents model must be used with following rules:
```
shilpa-padgaonkar, line 1315: For consistency across CAMARA APIs the uniform CloudEvents model must be used with following rules:

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1290,54 @@ response:
...
+| type | string | a value describing the type of event related to the originating occurrence. For consistency accross API we mandate following pattern: `org.camaraproject.<api-name>.<EVENT_NAME` (for exemple org.camaraproject.device-status.ROAMING_CHANGE_COUNTRY ) | mandatory |
+| specversion | string | version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version) | mandatory |
+| datacontenttype | string | media-type that describes the event payload encoding, must be `application/json` for CAMARA APIs| optional |
+| subject | string | describes the subject of the event - Not used in CAMARA notification. | optional |
+| time | string  datetime| timestamp of when the occurrence happened (must adhere on CAMARA datetime recommendation based on RFC 3339) | mandatory |
```
shilpa-padgaonkar, line 1325: For time, the note mentioned by CloudEvents spec document seems relevant as well. For some cases, it actually refers to when the event occured (for eg. in the case of QoD ) but in some cases, it is just about the current time reported by Event producers (for e.g device-status?).

https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time
May be worth adding this note here as well?

documentation/API-design-guidelines.md
```diff
@@ -1287,44 +1290,54 @@ response:
...
 | ----- |	-----  |	 -----  |  -----  | 
-| eventSubscriptionId | string | subscription identifier - could be valued for Resource-based subscription | optional |
-| event | object | event structure - see next table | mandatory |
+| id | string | identifier of this event, that must be unique in the source context. | mandatory |
+| source | string - URI | identifies the context in which an event happened in the specific Provider Implementation. Often this will include information such as the type of the event source, the organization publishing the event or the process that produced the event. The exact syntax and semantics behind the data encoded in the URI is defined by the event producer. | mandatory |
+| type | string | a value describing the type of event related to the originating occurrence. For consistency accross API we mandate following pattern: `org.camaraproject.<api-name>.<EVENT_NAME` (for exemple org.camaraproject.device-status.ROAMING_CHANGE_COUNTRY ) | mandatory |
+| specversion | string | version of the specification to which this event conforms (must be 1.0 if it conforms to cloudevents 1.0.2 version) | mandatory |
```
shilpa-padgaonkar, line 1322: Just as all Camara APIs conform to the same OAS version, we need to already fix the cloudevents spec version to the specific version number ( for e.g. 1.0.2) and not leave this up to the subprojects to choose on their own.

documentation/API-design-guidelines.md
```diff
@@ -1389,18 +1403,19 @@ curl -X 'POST' \
...
+  "data": {
+    "eventSubscriptionId": "456g899g",
+    "device": {
+      "phoneNumber": 123456789
+    },
+    "terminationReason": "SUBSCRIPTION_EXPIRED"
+  },
+  "time": "2023-01-19T13:18:23.682Z"
 }
 ```
```
shilpa-padgaonkar, line 1420: A section in the end on Security where we can point to https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#privacy-and-security

and also point that producers shall implement mechanisms at their end to ensure Abuse protection https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md#4-abuse-protection

shilpa-padgaonkar:
@bigludo7: Added some comments late yesterday evening.

rartych:
I am not completely certain about  using CAPITAL_AND_UNDERSCORE format for event type in `type` attribute.
It should be prefixed with a reverse-DNS name which is typically lower-case.

CloudEvent [spec](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#type) gives the following examples (only lower-case):
> 
 >    - com.github.pull_request.opened
 >    - com.example.object.deleted.v2

For uniformity I propose to use lower-case (and for  better human readability kebab-case as we already have in **API name** part:
`org.camaraproject.<api-name>.<event-name>`
or 
`org.camaraproject.<type>.<api-name>.<version>.<event-name>` if we apply versioning

For developers it may be important to use the most popular formatting as indicated here: https://github.com/camaraproject/Commonalities/pull/43#discussion_r1319631841

bigludo7:
> I am not completely certain about using CAPITAL_AND_UNDERSCORE format for event type in `type` attribute. It should be prefixed with a reverse-DNS name which is typically lower-case.
> 
> CloudEvent [spec](https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#type) gives the following examples (only lower-case):
> 
> > * com.github.pull_request.opened
> > * com.example.object.deleted.v2
> 
> For uniformity I propose to use lower-case (and for better human readability kebab-case as we already have in **API name** part: `org.camaraproject.<api-name>.<event-name>` or `org.camaraproject.<type>.<api-name>.<version>.<event-name>` if we apply versioning
> 
> For developers it may be important to use the most popular formatting as indicated here: [#43 (comment)](https://github.com/camaraproject/Commonalities/pull/43#discussion_r1319631841)

@rartych I'm fine with that - fair arguments for the example & uniformity
@jlurien as team Telefonica was pushing for CAPITAL_AND_UNDERSCORE could you be agreed on this?

documentation/API-design-guidelines.md
sachinvodafone: Sorry, I can't locate the "notification-as-cloud-event.yaml" file within the "commonalities/artifact" directory. I'm uncertain if I've overlooked something.

sachinvodafone: Is it possible for us to suggest a pattern for the "ID," whether it should be a UUID or another option?

rartych: @sachinvodafone  "notification-as-cloud-event.yaml" file is subject to another PR https://github.com/camaraproject/Commonalities/pull/43 .

Currently it is behind the changes we are discussing here in Design Guidelines.

documentation/API-design-guidelines.md
rartych: It looks that CloudEvents specs are pretty stable - since 2019 they advanced from 1.0.0 to 1.0.2. Since the format is implemented in multiple systems they should keep the backward compatibility as much as possible. We can safely stay with specversion "1.0" even the new version is introduced since here we use only base functionality.

documentation/API-design-guidelines.md
shilpa-padgaonkar: > Is it possible for us to suggest a pattern for the "ID," whether it should be a UUID or another option?
@sachinvodafone : Sure, could you kindly make a suggestion before the next commonalities, so that people have time to react to your suggestions before the meeting?

documentation/API-design-guidelines.md
bigludo7: @shilpa-padgaonkar 

Point 1: I do not think so. I think point raised by @PedroDiez in point 4 is about having or not subscription endpoint in same yaml than the direct response endpoint. 

Point 2: I need to understand - We provide in the yaml the definition of the notification (via the `callback`) - So for example in the Device Status API v2 you'll find the notification about Roaming status change.  With your proposal we introduce the API version in the type like `"type:org.camara.api.device-status.v2.RoamingStatusEvent`. I'm correct ?

bigludo7 (commented):
Taken into account @shilpa-padgaonkar review comments.

bigludo7 (commented):
Fixed some small miscues.

documentation/API-design-guidelines.md
bigludo7: @shilpa-padgaonkar Added 2 paragraphs to mention these point and link to CloudEvents specification.

documentation/API-design-guidelines.md
bigludo7: Fixed the version to V1.0.
... and in improbable case of new version occurs for CloudEvents I guess we'll revisit the paragraph.

documentation/API-design-guidelines.md
bigludo7: @shilpa-padgaonkar Yes you're right - done.

murthygorty (changes requested):
Hi, this is an excellent PR, thank you. A few key suggestions.

12.1. Subscription Types. 
Currently, I see that there is no query/cancel option for 'instance' subscription. 
I suggest an alternate approach: we treat the differences as simply "implicit"  and "explicit" - everything else is the same. Implicit ones get included as part of another request (QoD) and explicit ones are simply those, explicit subscriptions.
Both types have the same semantics wrt to query/cancel/update.

Secondly, for the 'notificationAuthToken', I think we SHOULD require that it be unique for every subscription. At the minimum, it should be HIGHLY RECOMMENDED, for security reasons, can elaborate if needed.


On the notifications themselves..
I fully endorse the use of a common model, CloudEvents seems as good as any. The spec for CloudEvennts.io itself is lacking in several aspects - quickstart guides, tutorials etc, requiring significant developer time investment. Do we have any pull in CNCF and get some of these implemented asap?
for CAMARA purposes, I really like that we have called out which of CloudEvents apply to us. We just need to include the library links in our 'implementations'.
cc: @gmuratk @shilpa-padgaonkar @hdamker

documentation/API-design-guidelines.md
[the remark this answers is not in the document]
patrice-conil: > The option is to use more readable kebab-case i.e. `org.camara.api.device-status.roaming-status-event`

Hi @PedroDiez, @rartych, @bigludo7 
Sorry, I'm not sure to understand  you well.
IMHO I thought here we are talking about the fully qualified class name and not the DNS FQDN and that is why we can use capital letters.
This is confusing because the package name mimics the reverse DNS name.
If I'm wrong, please feel free to correct me.

documentation/API-design-guidelines.md
jlurien: For Point 1, I don't see the necessity for the <type> part. The format is to be make event types universally unique, and under the scope of "org.camaraproject", I think that <api-name> is going to be unique. With that 2 different CAMARA APIs could define same <EVENT_NAME> but it wouldn't be a problem.

Regarding event versioning, if would be useful if we set that schema for internal data is linked to the event type, and in order to change the schema for data we have to update the event type, or at least when a backwards incompatible change happen.

bigludo7:
@murthygorty Thanks for the positive feedback

implicit/explicit: I like your proposal which is less confusing than instance-based or resource-based. If nobody object I can make the change (I tag the 'usual suspects' on the topic to get their view @jlurien @rartych @shilpa-padgaonkar @PedroDiez @patrice-conil ).

I propose also to add you comment on `notificationAuthToken` in the guideline

For the third point this is more a comment on globally the use of CloudEvents and I do not see a requirement to update the guideline - I'm right?

Again thanks for the review & the proposal.

documentation/API-design-guidelines.md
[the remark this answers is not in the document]
rartych: Please see my comment below: https://github.com/camaraproject/Commonalities/pull/56#issuecomment-1721337366

documentation/API-design-guidelines.md
shilpa-padgaonkar: For point 1: IF we simply assume the scope (of point 4)to be whether subscription yaml should be a separate file or not, you are correct, this is not needed. I am thinking more from a generic resource based subscription view, and in addition bringing the API families and later events under some structure. But may be this topic can also be asked in TSC & for now can be ignored here.

For point2: IF we consider resource based subscriptions, I don't believe that every event (in future) will have its source under some restful API like we have in the current cases. When we come to this point, it would be necessary to maintain the event versions to have visibility on the change in schema. 
Even though, this might be redundant as you have shown in your example for the current APIs in question, it will help keep the format consistent for all cases.

murthygorty:
> For the third point this is more a comment on globally the use of CloudEvents and I do not see a requirement to update the guideline - I'm right?

@bigludo7 - you are right in that guideline doesn't need to be updated. I'm taking the opportunity to request @shilpa-padgaonkar @hdamker and the likes to get CNCF improved perhaps via TechSteering so that OUR developers have an overall best experience, for us to succeed holistically :-) 

> [https://github.com/camaraproject/Commonalities/pull/56#issuecomment-1723420492v](url)

Hi @sachinvodafone, so sorry; couldn't fully understand if you are supportive of common approach to both subscription types or not. I really like your differentiation between subscriber and caller, if I understand you correctly. Are you suggesting that the use of the term 'implicit' brings confusion because that implies subscriber/consent involvement?

Personally, I'm not particular about the names, I'm however keen that we view both subscription-types the same in terms of the operations possible. And perhaps, since we are treating them the same largely, we don't even have to name them differently, thus simplifying the concept even further --> in some cases, subscription is 'automatic' as part of an operation; in some cases subscription is the operation itself, that's it. What do you say, @bigludo7 @shilpa-padgaonkar @hdamker

bigludo7 (commented):
Shifted event type to kebab-case

documentation/API-design-guidelines.md
```diff
@@ -1287,61 +1302,84 @@ response:
...
+| type | string | a value describing the type of event related to the originating occurrence. For consistency accross API we mandate following pattern: `org.camaraproject.<api-name>.<event-name>` (for exemple org.camaraproject.device-status.roaming-status ) | mandatory |
+| specversion | string | version of the specification to which this event conforms - must be "1.0" | mandatory |
+| datacontenttype | string | media-type that describes the event payload encoding, must be `application/json` for CAMARA APIs| optional |
+| subject | string | describes the subject of the event - Not used in CAMARA notification. | optional |
+| time | string  datetime| Timestamp of when the occurrence happened. If the time of the occurrence cannot be determined then this attribute MAY be set to some other time (such as the current time) by the CloudEvents producer, however all producers for the same `source` MUST be consistent in this respect. In other words, either they all use the actual time of the occurrence or they all use the same algorithm to determine the value used. (must adhere on CAMARA datetime recommendation based on RFC 3339) | mandatory |
```
jlurien, line 1339:
```suggestion
| time | string  datetime| Timestamp of when the occurrence happened. If the time of the occurrence cannot be determined then this attribute MAY be set to some other time (such as the current time) by the CloudEvents producer, however all producers for the same `source` MUST be consistent in this respect. In other words, either they all use the actual time of the occurrence or they all use the same algorithm to determine the value used. (must adhere on CAMARA datetime recommendation based on RFC 3339) | mandatory (*) |
```

documentation/API-design-guidelines.md
```diff
@@ -1287,61 +1302,84 @@ response:
...
+
+#### Security Considerations
+
+As notification may carry sensitive information, privacy and security constraints has to be considered. CloudEvents specification provides some guidance there: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#privacy-and-security
+
+#### Abuse Protection
+
+Any system that allows registration of and delivery of notifications to arbitrary HTTP endpoints can potentially be abused such that someone maliciously or inadvertently registers the address of a system that does not expect such requests and for which the registering party is not authorized to perform such a registration.
+
+In order to protect the sender, CloudEvents specification provides some guidance there: https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md#4-abuse-protection
```
shilpa-padgaonkar, line 1378: Event Producers shall choose based on their internal security guidelines to implement measures based on the above guidance to ensure abuse protection. For e.g. An event producer might ask the subscriber to pre-register the notification URL at the time of app onboarding. If this registered notification URL doesn't match later with the notification URL in the request, the event producer can choose to reject the request with the relevant error code.

bigludo7 (commented):
Added @shilpa-padgaonkar precision on security + introducing explicit/implicit wording

documentation/API-design-guidelines.md
bigludo7: Make sense to add this paragraph. I've added it. Thanks !

bigludo7:
From my perspective all significant comments impacting the introduction of event subscription & notification following CloudEvents format have been taken into account & resolved. 

I will provided this feedback during today TSC (@hdamker).

@rartych @shilpa-padgaonkar @jlurien @sachinvodafone @murthygorty please do not hesitate to jump in if there is still **blocking** point to implement this pattern in our CAMARA API (we can still in future PR improve the wording of the guideline).

documentation/API-design-guidelines.md
dfischer-tech, line 1: Small format adjustments would be appropriate in line:
216 - 218 : change "- - " to "--" or "  -"
and typos in lines
70 - "comunicate" 
120 - "analyse"
211 - "terminolgy"
217 - "mulitple"
288 - "senstive"
324 - "Unathourized"
405 - "recommendended"
412 - "Referer"
592 - "aAdded"
634 - "ellaborate" and "exisiing" and "errores"
709 - "returnss"
720 - "ascp" - i only know asc for sorting
1010 - "erros"
1050 - "desctiption"
1186 - "Anycase"
1324 - "explicited" (explicitly ?)
1335 - "accross" and "exemple"
1354 - "subscrition"
Thx

shilpa-padgaonkar:
@bigludo7  I agree that most important points are covered. 

However I would like to point out 2 last points :) :

- Even though resource based subscription will be dealt with later I found something in the documentation, For eg. we mention 4 operations under resource based subscriptions and they have paths as "/event-subscriptions" whereas the cloudevent subscription yaml file simply has "/subscriptions". IMHO unless we really need to differ, we can keep the things same? 

- Version:  How do we expect in resource based subscription the user to know the change in event schema if we dont include version in "type"  for eg. org.camaraproject.device-status.<version>.roaming-off?

bigludo7 (commented):
Following @shilpa-padgaonkar  and @akoshunyadi  :
Added `version `in the type `attribute`
Shifted endpoint from `/event-notifications` to `/notifications `

Thanks to @dfischer-tech  review fixed some miscues.

bigludo7:
Thanks @dfischer-tech  @shilpa-padgaonkar  and @akoshunyadi for your review !

@dfischer-tech As I worked on the notification part I have only fixed this part. Probably good to open an issue withe the identified adjustements.

sachinvodafone:
> Hi @sachinvodafone, so sorry; couldn't fully understand if you are supportive of common approach to both subscription types or not. I really like your differentiation between subscriber and caller, if I understand you correctly. Are you suggesting that the use of the term 'implicit' brings confusion because that implies subscriber/consent involvement?

HI @murthygorty , I was highlighting that the term "instance-based subscription," as described in our documentation, actually the definition of an "Implicit subscription" rather than an "instance-based subscription."

dfischer-tech:
@bigludo7: Thanks for the comment - issue is on the way.
https://github.com/camaraproject/Commonalities/issues/65

documentation/API-design-guidelines.md
```diff
@@ -1389,18 +1430,19 @@ curl -X 'POST' \
...
-      "ueId": {
-        "ipv4Addr": "192.168.0.1"
-      },
-      "uePort": 5060,
-      "terminationReason": "Service terminates for lack of consent"
-    }
-  }
+  "id": 123658,
+  "source": "https://notificationSendServer12.supertelco.com",
+  "type": "org.camara.api.device-status.v1.subcription-ends",
```
jlurien, line 1435:
```suggestion
  "type": "org.camaraproject.api.device-status.v1.subcription-ends",
```

documentation/API-design-guidelines.md
```diff
@@ -1183,16 +1198,16 @@ In order to ease developer adoption, the pattern for Resource-based event subscr
...
+| DELETE | `/subscriptions/{subscriptionsId}` | Operation to delete a event subscription |
 
 
 Note on the operation path:
-The recommended pattern is to use `/event-subscriptions` path for the subscription operation. But API design team, for specific case, has the option to append `/event-subscriptions` path with a prefix (e.g. `/roaming/event-subscriptions` and `/connectivity/event-subscriptions`). The rationale for using this alternate pattern should be explicitly provided (e.g. the notification source for each of the supported events may be completely different, in which case separating the implementations is beneficial). 
+The recommended pattern is to use `/subscriptions` path for the subscription operation. But API design team, for specific case, has the option to append `/subscriptions` path with a prefix (e.g. `/roaming/subscriptions` and `/connectivity/subscriptions`). The rationale for using this alternate pattern should be explicitly provided (e.g. the notification source for each of the supported events may be completely different, in which case separating the implementations is beneficial).
```
shilpa-padgaonkar, line 1208: Wouldn't this kind of an (optional) categorization or extended information be more better in the type `org.camaraproject.<optional-category or extension><api-name>.<api-version>.<event-name>`rather than in the path? WDYT?

Other than that I am almost ready to approve this review :)

bigludo7: @shilpa-padgaonkar Not sure to follow you. You propose to update the `type` attribute with an additional `<optional-category or extension>` but this point in the guideline is not about the event type but the `/subscriptions` endpoint. 

What we say here is for explicit subscription pattern we recommend a generic `/subscriptions` endpoint but for some API with several capabilities we allow to 'specialize' this subscription endpoint (and provide several) to make it very explicit for the developer.

shilpa-padgaonkar: @bigludo7 Then may be the examples were misleading to me :)  Because the examples shown here at least to me seem as some sort of additional info /categorization about the event. If this is not the case, you can ignore my comment :)

bigludo7: Thanks @shilpa-padgaonkar  - indeed this is not related to the event directly but to the subscription.

shilpa-padgaonkar (dismissed):
/LGTM

documentation/API-design-guidelines.md
```diff
@@ -1183,16 +1198,16 @@ In order to ease developer adoption, the pattern for Resource-based event subscr
 
 | operation | path | description |
 | ----- |	-----  |	 -----  | 
-| POST | `/event-subscriptions` |  Operation to request a event subscription.     |
-| GET | `/event-subscriptions` |  Operation to retrieve a list of event subscriptions - could be an empty list.  eg. `GET /event-subscriptions?type=ROAMING_STATUS&ExpireTime.lt=2023-03-17` |
-| GET | `/event-subscriptions/{eventSubscriptionsId}` | Operation to retrieve a event subscription |
-| DELETE | `/event-subscriptions/{eventSubscriptionsId}` | Operation to delete a event subscription |
+| POST | `/subscriptions` |  Operation to request a event subscription.     |
+| GET | `/subscriptions` |  Operation to retrieve a list of event subscriptions - could be an empty list.  eg. `GET /subscriptions?type=ROAMING_STATUS&ExpireTime.lt=2023-03-17` |
```
rkandoi, line 1202: The type field value here probably needs to align with the new convention?

rkandoi:
One of the goals of the PR is "_enforce use of callbacks in CAMARA API to describe notification event structure_". Im not sure if that been covered. If it has, could you kindly point to the line numbers in the commit. Thank you in advance.

documentation/API-design-guidelines.md
bigludo7: Agreed - will fix.

bigludo7 (commented):
Thanks @rkandoi for the review
Fixed the 2 points you raised.

documentation/API-design-guidelines.md
```diff
@@ -1183,16 +1198,16 @@ In order to ease developer adoption, the pattern for Resource-based event subscr
 
 | operation | path | description |
 | ----- |	-----  |	 -----  | 
-| POST | `/event-subscriptions` |  Operation to request a event subscription.     |
-| GET | `/event-subscriptions` |  Operation to retrieve a list of event subscriptions - could be an empty list.  eg. `GET /event-subscriptions?type=ROAMING_STATUS&ExpireTime.lt=2023-03-17` |
```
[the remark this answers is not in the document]
rartych: @rkandoi For operations such as  "lt" there is still open issue: https://github.com/camaraproject/Commonalities/issues/10

bigludo7 (commented):
Renamed attribute (and aligned in the example) `eventSubscriptionId` to `subscriptionId` to be aligned with endpoint name.

documentation/API-design-guidelines.md
rartych: On one side alignment of subscription `type` and notification `type` is good for uniformity, but on the other side only notifications are covered by CloudEvents specification and requirements and using the reverse DNS notation for event subscription looks strange for me.
Using old simple notation in subscription makes another feature more intuitive: within one subscription events of different types can be generated, for example: Subscription to `type=ROAMING_STATUS` can result in `org.camaraproject.device-status.v1.roaming-on` or `org.camaraproject.device-status.v1.roaming-off` events.
For clarity we can distinguish parameter names and have subscription `kind`  and notification `type` as required by CloudEvents.
What are opinions of others?

patrice-conil: Hi @rartych, 
I agree with you, using "kind" in subscription might be more user friendly than using event type.
Of course we need more opinions.

documentation/API-design-guidelines.md
```diff
@@ -1138,23 +1152,23 @@ Managed event types are explicitly defined in CAMARA API OAS.
 ### 12.1 Subscription
 
 We distinguish 2 types of subscriptions:
-- Instance-based subscription (indirect creation)
-- Resource-based subscription (direct creation)
+- Instance-based subscription (implicit creation)
+- Resource-based subscription (explicit creation)
 
-**Instance-based subscription**
+#### Instance-based (implicit) subscription
```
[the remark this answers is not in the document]
bigludo7: @sachinvodafone Sorry i did not understand your question - my I ask you to elaborate? Thanks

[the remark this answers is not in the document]
sachinvodafone: Hi @bigludo7 , Just to clarify, I understand that an 'instance-based' subscription is associated with a specific event or instance of resource creation, whereas an 'implicit subscription' suggests that the subscription is established indirectly or automatically due to some other action or event.

Similarly, we also have two other distinct subscription types: 'resource-based' and 'explicit' subscriptions. In this context, we could potentially have four combinations: 

- 'Instance-based subscriptions (explicit),' 
- 'Instance-based subscriptions (implicit),' 
- 'Resource-based subscriptions (explicit),' 
- and 'Resource-based subscriptions (implicit).'

However, our documentation currently only covers two of these combinations:
* 'Instance-based subscription (implicit creation)'
* 'Resource-based subscription (explicit creation)'

shilpa-padgaonkar (approved):
/LGTM

patrice-conil (approved):
/LGTM

documentation/API-design-guidelines.md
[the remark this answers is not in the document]
bigludo7: @sachinvodafone Got it ! thanks for the clarification
Instance-based and implicit are same thing, and resource-based and explicit are same; As suggested by @murthygorty I have introduced this vocabulary to explain the 2 patterns. I feel sorry it creates confusion. I could update it in next release if other people thinks that it adds more confusion than clarification.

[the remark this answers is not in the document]
shilpa-padgaonkar: @sachinvodafone : If adding the above said description is not a blocking issue for you, could you please approve this PR as you were active in this review and kindly open a new issue for improving this documentation? Would that work for you?

murthygorty:
[like]  Gorty, Murthy reacted to your message:
________________________________

rartych (approved):

sachinvodafone (approved):

shilpa-padgaonkar:
@jlurien : Any blocking issues from your side? IF not, could you kindly approve, as you were an active reviewer?
Once your approval is done, I believe we can merge this PR.

documentation/API-design-guidelines.md
PedroDiez: According to CloudEvents spec, the "type" value needs to be coherent among events and related subscriptions. Then, it is needed to have the same value and follow DNS-reverse format with the CAMARA pattern.

Reference:
https://github.com/cloudevents/spec/blob/main/subscriptions/spec.md#321-subscription-object

> Type: Array of Strings - array of CloudEvents type values

Description: Indicates which types of events the subscriber is interested in receiving. _When present on a subscribe request, all events generated due to this subscription MUST have a CloudEvents type property that matches one of these values._

Constraints:

OPTIONAL
If present, any value present MUST a non-empty string
Examples:

com.github.pull_request.opened
com.example.object.deleted

PedroDiez (approved):
Some editorial comments.

LGTM with the current status

shilpa-padgaonkar:
As most active reviewers have approved the PR and confirmed that there are no more blocking issues, will merge this PR. Thank you all for the great collaboration on this PR and a big thanks to @bigludo7 for driving this topic .

Issue #46, opened 2023-08-07
Using kebab-case for security scope is not easily machine readable


RandyLevensalor:
**Problem description**
In the [Security Implementation](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#102-security-implementation) it suggests using kebab-case for the scope.

The provided example `explore-content-modules-users-read` is not easily machine readable.

**Expected action**
Ideally, a developer would be able to use the .split('-') operation on the scope and get three strings in return.  One for each of the three elements.  <API_name>-<Resource>-<Permission>.  Since the elements are also kebab-case, the split command would return 5 strings from the split operation.  And then would need to have logic that would know what the expected API name and resource name are.

If there was one api named explore and another named explore-modules, the code would need to be aware of this and couldn't use a shared library for parsing the the permissions.

Using snake case or camel case for the three elements and kebab-case for combining the elements would resolve this.

**Additional context**
To date, all of the APIs have a single name, so identifying the API name or updating the mechanism the the API name shouldn't impact any current APIs.

It's also unclear if a multi word api name should be kebab-case or another format.  Section 4.1 seems to hint that it should be kebab-case.

> URI with lowercase and hyphens. URIs must be "human readable" to facilitate identification of the offered resources. Lowercase words and hyphenation (kebab-case) help achieve this best practice. For example: /customer-segments

These may not be items that there is agreement to change, but at the very least it would be appreciated to make the documentation explicit on these topics.

bigludo7:
@RandyLevensalor For the device location API, @jlurien proposed to use format `api_name:[resource:]action` for scope in security scheme (use of `:` as separator). Is it something working for you and that we can propose for all APIs?

There is also an open issue on same topic in Number verify (https://github.com/camaraproject/NumberVerification/issues/49)

jlurien:
We are preparing a proposal for common guidelines around scopes naming and security/securitySchemes specification. We'll open a new issue for this.

jlurien:
New issue created: https://github.com/camaraproject/Commonalities/issues/53

PR #57, opened 2023-09-06
Update OpenAPI security guidelines.md

### LINKED ISSUE #53: Unify specification of `securitySchemes`, `security` and scopes
## **Problem description**

Currently, there is no consistency across WG regarding definition of the `securitySchemes` types, and about providing and naming scopes to protect operations. 

This table summarizes the situation, listing which `securitySchemes` are considered per WG, and which scopes are defined for that type (linked by means of `security` property):

| WG | oauth2/<br>clientcredentials | oauth2/<br>authorizationCode | openIdConnect |
| :--- | :--- | :--- | :--- |
| AnonymisedSubscriberIdentifier | Without scopes | - | - |
| BlockChainPublicAddress (PR) | - | `blockchain-public-address-read`<br>`blockchain-public-address-create`<br>`blockchain-public-address-delete` | - | - |
| CarrierBillingCheckOut | Without scopes | `carrier-billing-payments-create`<br>`carrier-billing-payments-prepare`<br>`carrier-billing-payments-confirm`<br>`carrier-billing-payments-cancel`<br>`carrier-billing-payments-read` | - |
| DeviceIdentifier | Without scopes | - | - |
| DeviceLocation | Without scopes | `location-verification:verify` | - |
| DeviceStatus | Without scopes | `device-status-roaming-read`<br>`device-status-roaming-event` | - |
| EdgeCloud | `nbi-mgmt`<br>`edge:discovery:read`<br>`edge:serviceprofile:read`<br>`edge:serviceprofile:write`<br>`edge:serviceregistry:read`<br>`edge:serviceregistry:write`<br>`edge:traffic-influences:read`<br>`edge:traffic-influences:write` | - | - |
| HomeDevicesQoD | - | - | `home-devices-qod-qos-write` |
| NumberVerification (CAMARA) | - | - | `number-verification-verify-read`<br>`number-verification-share-read` |
| OTPValidationAPI | `one-time-password-sms:send-validate` | - | - |
| SimSwap (CAMARA) | Without scopes | `retrieve-sim-swap-date`<br>`check-sim-swap` | - |
| QualityOnDemand | Without scopes | `qod-sessions-read`<br>`qod-sessions-write`<br>`qod-sessions-delete`<br>`qod-profiles-read` | - |

## **Possible evolution**
Unify specification of `securitySchemes`, `security` and scopes across all WGs with common guidelines. Exceptions to the guidelines should be avoided and justified.

### Use of openIdConnect for `securitySchemes`

Generally, OpenID Connect is the protocol to be used for securitization. Each API spec will define the following entry in `securitySchemes`, as shown in [CAMARA-AuthN-AuthZ-Concept.md](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-AuthN-AuthZ-Concept.md#documentation-and-specs-) document:
```
components:
  securitySchemes:
    openId:
      type: openIdConnect
      openIdConnectUrl: https://example.com/.well-known/openid-configuration
```

The value for `openIdConnectUrl` in the CAMARA spec is an example, that must be substituted by the specific discovery endpoint for OIDC protocol of the API provider, when the API is exposed in one of its environments.

### Use of `security` property

Generally, each operation will be protected by a scope and it will include a `security` property with a single element in the array:

```
paths:
  {path}:
    {method}:
      ...
      security:
        - openId:
            - {scope}
```

The key is arbitrary in OAS, but convention in CAMARA is to name it `openId`. This key must be same defined in the `components.securitySchemes` section.
The {scope} is the specific scope defined to protect this operation. It must follow the syntax specify below.

### Scope naming

Define a scope per API operation with the structure:

`api-name:[resource:]action`

where

* `api-name` is the API name specified as the base path, prior to the API version, in the `servers[*].url` property. For example, from `/location-verification/v0`, it would be `location-verification`.

* `resource` is optional. For APIs with several `paths`, it may include the resource in the path. For example, from `/qod/v0/sessions/{sessionId}`, it would be `sessions`.

* `action`: There are two cases:
  - For POST operations with a verb, it will be the verb. For example, from `POST /location-verification/v0/verify`, it would be `verify`.
  - For CRUD operations on a resource in paths, it will be one of:
    - `create`: For operations creating the resource, typically `POST`.
    - `read`: For operations accessing to details of the resource, typically `GET`.
    - `update`: For operations modifying the resource, typically `PUT` or `PATCH`.
    - `delete`: For operations removing the resource, typically `DELETE`.
    - `write` : For operations creating or modifying the resource, when differentiation between `create` and `update` is not neeed.

#### Examples

| API | path | method | scope |
| --- | --- | --- | --- |
| location-verification | /verify | POST | `location-verification:verify` |
| qod | /sessions | POST | `qod:sessions:create`, or<br>`qod:sessions:write` |
| qod | /qos-profiles | GET | `qod:qos-profiles:read` |

jlurien:
#### What type of PR is this?

* documentation

Add detailed guidelines on how to model scopes, security and securitySchemes in OpenAPI spec.

* Scope format
* openIdConnect as preferred securityScheme type 

Based on discussion and comments - resolves #53
It also fixes https://github.com/camaraproject/Commonalities/issues/46 

#### Changelog input

```
 Changes section 11.6 Security definition
```

#### Additional documentation 

n/a


eric-murray:
Please review or comment by next Commonalities meeting (currently 21st, but may be moved due to clash with next TSC meeting).

jordonezlucena (dismissed):
LGTM

documentation/API-design-guidelines.md
```diff
@@ -1116,17 +1101,73 @@ When IpAddr is used in a payload the property objectType MUST be present to indi
...
+        - openId:
+            - {scope}
+```
 
-- Security Type: oauth2, oauth…
-- Security Flow (Depends of security type): implicit, password…
-- Security Flow description applied (String)
-- Endpoint token URL
-- URL to endpoint authorization ( If flow is based on "`authorizationCode`").	
+The key is arbitrary in OAS, but convention in CAMARA is to name it `openId`. This key must be same defined in the `components.securitySchemes` section.
```
shilpa-padgaonkar, line 1169: Needs to align with the review of https://github.com/camaraproject/IdentityAndConsentManagement/pull/62

documentation/API-design-guidelines.md
```diff
@@ -1116,17 +1101,73 @@ When IpAddr is used in a payload the property objectType MUST be present to indi
 }
 ```
 
-### 11.6 OAuth Definition
+### 11.6 Security definition
+
+Finally, this part describes the security is applied to the API.
```
shilpa-padgaonkar, line 1106: This section describes and aligns the security specific definitions within the API specification, such as scopes and the security schemes.

jlurien, line 1106:
```suggestion
This section describes and aligns the security specific definitions within the API specification, such as scopes and the security schemes.
```

jlurien: Updated

documentation/API-design-guidelines.md
```diff
@@ -1116,17 +1101,73 @@ When IpAddr is used in a payload the property objectType MUST be present to indi
...
+| API | path | method | scope |
+| --- | --- | --- | --- |
+| location-verification | /verify | POST | `location-verification:verify` |
+| qod | /sessions | POST | `qod:sessions:create`, or<br>`qod:sessions:write` |
+| qod | /qos-profiles | GET | `qod:qos-profiles:read` |
+
+
+#### 11.6.2 Use of openIdConnect for `securitySchemes`
+
+Generally, OpenID Connect is the protocol to be used for securitization. Each API spec will define the following entry in `securitySchemes`, as shown in [CAMARA-AuthN-AuthZ-Concept.md](https://github.com/camaraproject/IdentityAndConsentManagement/tree/main/documentation/CAMARA-AuthN-AuthZ-Concept.md) document:
```
eric-murray, line 1143: I think "Generally" is doing too much work here. At least a link to the Identity & Consent documentation should be given, if not explicitly listing the conditions under which OIDC must be used.

documentation/API-design-guidelines.md
```diff
@@ -1116,17 +1101,73 @@ When IpAddr is used in a payload the property objectType MUST be present to indi
...
+##### Examples
+
+| API | path | method | scope |
+| --- | --- | --- | --- |
+| location-verification | /verify | POST | `location-verification:verify` |
+| qod | /sessions | POST | `qod:sessions:create`, or<br>`qod:sessions:write` |
+| qod | /qos-profiles | GET | `qod:qos-profiles:read` |
+
+
+#### 11.6.2 Use of openIdConnect for `securitySchemes`
```
eric-murray, line 1141: `openIdConnect` is not the only possibility for security schemes. `clientCredentials` flow is also allowed under certain circumstances. A template for the client credentials flow should also be provided in this document for completeness.

documentation/API-design-guidelines.md
```diff
@@ -1116,17 +1101,73 @@ When IpAddr is used in a payload the property objectType MUST be present to indi
...
+    openId:
+      type: openIdConnect
+      openIdConnectUrl: https://example.com/.well-known/openid-configuration
+```
+
+The value for `openIdConnectUrl` in the CAMARA spec is an example, that must be substituted by the specific discovery endpoint for OIDC protocol of the API provider, when the API is exposed in one of its environments.
+
+#### 11.6.3 Use of `security` property
+
+Generally, each operation will be protected by a scope and it will include a `security` property with a single element in the array:
```
eric-murray, line 1157: I don't think "Generally" is required here. The rule should be that every operation should be protected by a scope. If there are exceptions, they should be explicitly listed.

An example for client credentials should also be given.

bigludo7: @eric-murray I will formulate this way "As soon as user data are manipulated by an operation, the operation must be protected by a scope".

documentation/API-design-guidelines.md
bigludo7: @jlurien @eric-murray I understood that this proposal does not prevent to use `clientCredentials`;
With this proposal we extract the authorization information from the yaml. The authorization are described in another json that can be found here: https://example.com/.well-known/openid-configuration (as described here: https://swagger.io/docs/specification/authentication/openid-connect-discovery/)

Did I miss something?

eric-murray: So the decision is to use `openIdConnect` as the only possible security scheme?

I guess at least that simplifies the API definition at the cost of not knowing what authentication schemes are supported. The design guidelines should at least contain a note to this effect, because using `openIdConnect` as a way of specifying client credentials is to be used is not particularly intuitive.

documentation/API-design-guidelines.md
eric-murray: So the proposal is to use `openIdConnect` but with no scope when user data is not involved? Something like:
```
  security:
  - openid: []
```

I would clarify that and formulate it the other way around:
"The operation must be protected by a scope, unless no user data will be manipulated by the operation in which case no scope need be specified for the `openIdConnect` security scheme."

An example of specifying no scope might also be useful.

documentation/API-design-guidelines.md
eric-murray: One more question - would the API definitions that currently call the security scheme of type `openIdConnect` "`three-legged`" be expected to change that to "`openId`"? So rather than have two schemes of type `openIdConnect` called "`three-legged`" and "`two-legged`", we have just one called "`openId`" that covers both cases?

jlurien: This issue was opened prior to the later discussion in I&CM WG about this topic, and the debate is going on there, so in the end we should align Commonalities guidelines with the decision taken there. We have to review how to consolidate relevant information with details in one single place.

Which auth flow is allowed to be used for certain API will depend on many considerations that may change from country to country, that may depend on the business segment, on the considerations about which Personal Information is affected, etc. The idea is of this proposal is to make CAMARA API specifications the more "agnostic" as possible in relation to these concepts.

We are somehow limited here by the semantics in OAS. The ultimate requirement to me is to standardize the scopes which protect the different operations, but for that OAS requires to link the scopes definition in `security` to some entry in the `components/securitySchemes`  section by means of a key. The key itself is free for us to decide. We may choose "openId" or whatever other term, such as something even more generic, such as "scopes", etc. But the entry in securityScheme must be one of the allowed: oauth2, openIdConnect, etc. For sure we may have several entries in security and duplicate the scopes per entry, but using openIdConnect with .well-known discovery URL, which may allow client_credentials if necessary, was to me the most simple option.

documentation/API-design-guidelines.md
jlurien: I agree that "Generally" must be avoid and give precise guidelines. In OAS we can define a security entry at API root level, which applies to all operations which do not overwrite security. In this sense, it will be better to add what @eric-murray  suggests to a generic root section. Or we may agree that all operations must be protected by some scope, which I think is a good practice and what is commonly happening already in all of our specs.

bigludo7: @jlurien @eric-murray I'm fine with Eric proposal: "_The operation must be protected by a scope, unless no user data will be manipulated by the operation in which case no scope need be specified for the openIdConnect security scheme._" 

If we provide for example API of UE density on a geo zone, we haven't user data here and so we did not need a scope - correct?

eric-murray: @bigludo7
My concern with allowing "no scope" is that we might think there is no user data involved, but later find that privacy / legal teams say there is. Always having a scope for each operation, even if we think granting this scope will be automatic, allows for additional checks to be added later. I don't see that this complicates things particularly, but I&C sub-project would know better than me.

As an example, the use case you mention (UE density) is coming via GSMA, and is one Vodafone have been looking at. The view of our privacy teams was that this data was "analytics", and hence any subscribers included in the count would have to have opted-in for that.

Now, of course, in this case it would be ridiculous to check which subscribers had consented and calculate the UE density dynamically on that basis for every request for UE density information. The data would just be updated periodically taking this consent into account, but could then be shared freely. However, I give this example to show that it is not always obvious whether an operation is "doing something" with user data that requires consent.

documentation/API-design-guidelines.md
eric-murray: Yes, I agree that a single `openId` scheme which says "look in `./well-known/openid-configuration` for details" is the most flexible, but is just moving the problem of standardising security schemes across implementations elsewhere.

I would guess we will end up defining a template `openid-configuration` file for each API definition (or multiple files, depending on use case or country) showing what is allowed. Or maybe such details will only be in the documentation.

jlurien: We may agree also in adding some section in info.description for "# Authentication and authorization", explaining how security and securitySchemes should be interpreted.

documentation/API-design-guidelines.md
jlurien: Agree also, I think it is good practice to add scopes to all operations. As legal and regulatory considerations are complex and difficult to predict, it is safer to avoid making assumptions while defining the APIs

gregory1g: @eric-murray , always having a scope sounds good with an option to grant it automatically sounds like a good way forward.

I suggest to avoid "user data are manipulated by an operation" - this allows different interpretations. One can believe that "manipulation" means "changing". then reading user's location is not a "manipulation".

shilpa-padgaonkar (approved):
/LGTM

documentation/API-design-guidelines.md
jlurien: This part will be removed and referenced to the document in I&CM

documentation/API-design-guidelines.md
jlurien: This section is duplicating the work in I&CM, so I will reference the work there

documentation/API-design-guidelines.md
jlurien: This section is duplicating the work in I&CM, so I will reference the work there

PedroDiez (approved):
LGTM

rartych (approved):

patrice-conil (approved):
LGTM

PR #82, opened 2023-10-30
Filtering GET results - update section 8.3 in API DG 

### LINKED ISSUE #10: Attribute for filtering in GET request

rartych:
Clarifications and example of comparison attributes in 8.3 Filtering

#### What type of PR is this?

* documentation

Clarification and example how to use filtering attributes using comparisons like 'greater or equal' 

Fixes #10 

Modified original examples as  not clear

#### Changelog input

```
Clarification for filtering attributes of GET request using comparisons like 'greater or equal'  added

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -744,7 +744,7 @@ Next, it is specified how it should be used according to the filtering based on 
 **Additional rules**:
 - The operator "`&`" is evaluated as an AND between different attributes.
 - A Query Param (attribute) can contain 1 or n values separated by "`,`".
-- For operations on numeric, date or enumerated fields, the use of the suffixes `.(gte|gt|lte|lt)$` will be allowed, which will act as comparators for “greater - equal to, greater than, smaller - equal to, smaller than”.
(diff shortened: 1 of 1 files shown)


RubenBG7:
Here if we read the API guidelines designed, we are applying operators gt, lt etc to enums too, that means that we are allowing to compare strings with lt, gt.... that not make sense.

In adition we are not allowing with the actual guidelines to compare a value that is exactly that value , we have not defined the equals operator. Then my proposal is basing on the TMF 630 use:

image

For numeric and dates value allow:

.gt > /  .gte >= /  .lt < /  .lte <= /  .eq = / .neq !=

And for string enums:

 .eq = / .neq !=

patrice-conil (approved):
/LGTM

RubenBG7:
"For operations on numeric, date or enumerated fields, the parameters with the suffixes .(gte|gt|lte|lt)$ need to be defined, which should be used as comparators for “greater - equal to, greater than, smaller - equal to, smaller than” respectively. Only the parameters needed for given field should be defined e.g. with .gte and .lte suffixes only."

As I exposed , one string cannot be gt o lt than other. My comments was not take into account.

PR #115, opened 2023-12-20
Set X-Correlator as Required in OAS Definition, X-Version removed

### LINKED ISSUE #101: Should the headers like x-correlator be included in OAS definition?
**Problem description**
Proble occured in conversation to PR https://github.com/camaraproject/Commonalities/pull/88
Currently the table in API Design Guidelines "Required in OAS Definition" states "No" for both "X-Correlator" and "X-Version"
`X-Correlato`r header is included in some CAMARA API definitions.

In section 3.5 in relation to security related headers guidelines are:

> To avoid cluttering the CAMARA OAS (Swagger) definition files, the above headers must not be included explicitly in the API definition files even when supported, as it can be assumed that developers will already be familiar with their use.

**Expected action**
 The column "Required in OAS Definition" should rather be "Included in OAS Definition" and binding for all CAMARA APIs.
Or the whole section 9. Architecture Headers can be integrated within section 3.5.

**Additional context**
The more common names for header used to correlate HTTP requests between a client and server are `X-Correlation-ID` or `Correlation-ID`
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

jlurien:
#### What type of PR is this?

* documentation

- Removed X-Version
- X-Correlator is now set as "Required in OAS Definition"

Fixes #101

### DIFF
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -814,8 +814,7 @@ With the aim of standardizing the request observability and traceability process
 
 | Name | Description |  Type | Pattern	| Longitude | Location | Required by API Consumer | Required in OAS Definition |	Example | 
 |---|---|---|---|---|---|---|---|---|
-| `X-Version` |	Service version description to help observability process |	String| N/A	| | Request | No | No | |	
-| `X-Correlator`|	Service correlator to make E2E observability |		String |	UUID (8-4-4-4-12)	| Max 36	| Request/Response | No | No |	b4333c46-49c0-4f62-80d7-f0ef930f1c46 |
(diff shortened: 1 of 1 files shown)


RubenBG7 (approved):

uwerauschenbach:
Hi, a question: https://spec.openapis.org/oas/v3.1.0 does not mention X-Correlator. Where is it stated in OAS that it is required?

jlurien:
> Hi, a question: https://spec.openapis.org/oas/v3.1.0 does not mention X-Correlator. Where is it stated in OAS that it is required?

We are not meaning here that it is required by the OAS standard to be included, we mean that CAMARA requires that this header is documented in our OAS specs, so the resulting API specs are consistent.

patrice-conil (approved):
/LGTM

shilpa-padgaonkar (approved):
/LGTM

PR #120, opened 2024-01-10
Update CAMARA_common.yaml: add Device headers, fix address

### LINKED ISSUE #118: CAMARA Common YAML: addition of common HTTP Headers for Device Identifiers
**Problem description**
[CAMARA_common.yaml](https://github.com/camaraproject/Commonalities/edit/main/artifacts/CAMARA_common.yaml) does not account for Device Identifiers being provided as HTTP headers.  Use of HTTP Headers is a recommendation in the [API guidelines -  
POST or GET for transferring sensitive or complex data](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#post-or-get-for-transferring-sensitive-or-complex-data):

"When using GET, transfer the data using headers, which are not routinely logged or cached"

**Possible evolution**
Add reusable parameter definitions for these HTTP headers in [CAMARA_common.yaml](https://github.com/camaraproject/Commonalities/edit/main/artifacts/CAMARA_common.yaml) 

Headers are typically Uppercase/hyphen-separated, and are documented that way in the [HTTP Headers Definition section](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#35-http-headers-definition) of the guidelines.  Hence Device Identifier HTTP Headers would follow this convention:

`IPv4-Address`
`IPv6-Address`
`Network-Access-Identifier`
`Phone-Number` 

**Alternative solution**
Define Device HTTP headers in each relevant CAMARA spec, which risks misalignment.

**Additional context**

### LINKED ISSUE #119: CAMARA Common YAML: Both Address and Addr are used in object names
**Problem description**
[Common.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml) uses 'Addr' in the object name for an IPv4 address, but (correctly) uses 'Address' in other object names:

Use of Addr:
`#/components/schemas/SingleIpv4Addr`
`#/components/schemas/DeviceIpv4Addr`

Use of Address:
`ipv4Address`
`ipv6Address`
`#/components/schemas/DeviceIpv6Address`
` publicAddress:`
`privateAddress`

The full form ("address" or "Address") is used throughout the documentation, while Addr is not.

**Expected behavior**
Change `Addr` to `Address`

**Alternative solution**

**Additional context**

Kevsy:
Fixes #118  #119

#### What type of PR is this?

* correction
* enhancement/feature

- Adds HTTP Headers for Device Identifiers, for reuse in APIs with GET methods requiring protection for Device Identifiers
- Changes instances of `Addr` to `Address` in object names.

Fixes #118 
Fixes #119

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -10,6 +10,31 @@ info:
  version: 0.4.0
 paths: {}
 components:
+ parameters:
+  IPv4-Address:
+    in: header
+    name: IPv4-Address
+    required: false
+    schema:
+      $ref: "#/components/schemas/DeviceIpv4Address"
+  IPv6-Address:
+    in: header
+    name: IPv6-Address
+    required: false
+    schema:
+      $ref: "#/components/schemas/DeviceIpv6Address"
+  Phone-Number:
+    in: header
+    name: Phone-Number
+    required: false
+    schema:
+      $ref: "#/components/schemas/PhoneNumber"     
+  Network-Access-Identifier:
+    in: header
+    name: Network-Access-Identifier
+    required: false
+    schema:
+      $ref: "#/components/schemas/NetworkAccessIdentifier"
  schemas:
   TimePeriod:
    properties:
@@ -55,7 +80,7 @@ components:
     networkAccessIdentifier:
      $ref: "#/components/schemas/NetworkAccessIdentifier"
     ipv4Address:
-     $ref: "#/components/schemas/DeviceIpv4Addr"
+     $ref: "#/components/schemas/DeviceIpv4Address"
     ipv6Address:
      $ref: "#/components/schemas/DeviceIpv6Address"
    minProperties: 1
@@ -71,7 +96,7 @@ components:
    type: string
    example: "123456789@domain.com"
 
-  DeviceIpv4Addr:
+  DeviceIpv4Address:
    type: object
    description: |
     The device should be identified by either the public (observed) IP address and port as seen by the application server, or the private (local) and any public (observed) IP addresses in use by the device (this information can be obtained by various means, for example from some DNS servers).
@@ -83,9 +108,9 @@ components:
     In all cases, publicAddress must be specified, along with at least one of either privateAddress or publicPort, dependent upon which is known. In general, mobile devices cannot be identified by their public IPv4 address alone.
    properties:
     publicAddress:
-     $ref: "#/components/schemas/SingleIpv4Addr"
+     $ref: "#/components/schemas/SingleIpv4Address"
     privateAddress:
-     $ref: "#/components/schemas/SingleIpv4Addr"
+     $ref: "#/components/schemas/SingleIpv4Address"
     publicPort:
      $ref: "#/components/schemas/Port"
    anyOf:
@@ -95,7 +120,7 @@ components:
     publicAddress: "84.125.93.10"
     publicPort: 59765
 
-  SingleIpv4Addr:
+  SingleIpv4Address:
    description: A single IPv4 address with no subnet mask
    type: string
    format: ipv4


patrice-conil (commented):
LGTM 
But as this will impact many existing APIs (as IP addresses are often used) we need to ask other companies for their opinions.

artifacts/CAMARA_common.yaml
```diff
@@ -10,6 +10,31 @@ info:
  version: 0.4.0
 paths: {}
 components:
+ parameters:
+  IPv4-Address:
```
jlurien, line 14: Why not using same format for all schemas? If we use UpperCamelCase for schema names, it is more coherent to use them as well here, e.g. Ipv4Address

jlurien:
> Thank you @patrice-conil for pointing this out. As this small change: ` Addr` to `Address ` impacts subprojects like Device Identifier, Device Status and Device Location, QoD I am adding @eric-murray @bigludo7 @jlurien and @hdamker as reviewers.
> 
> Note: for IPv6 the `DeviceIpv6Address` name is used.

The change affects code generation as those names are schema names that may be mapped to classes. The API itself does not change.

eric-murray (approved):

jlurien (changes requested):
- Adopt UpperCamelCase for parameters schemas

artifacts/CAMARA_common.yaml
Kevsy: By convention HTTP headers are Kebab-Case:  [RFC 2616, section 5.3](https://www.rfc-editor.org/rfc/rfc2616#section-5.3).  So `IPv4-Address` is the appropriate **parameter name** for an HTTP header, whilst  
"DeviceIpv4Address" is the schema that the parameter is based on (which follows CAMARA UpperCamelCase convention for schema names).

Kevsy:
> * Adopt UpperCamelCase for parameters schemas

The **schemas** for the new parameters do have UpperCamelCase names.

But the **parameter names** (that reference the schemas) have Kebab-Case names, because they are specifically  HTTP header parameters, which follow the convention in [RFC 2616, section 5.3](https://www.rfc-editor.org/rfc/rfc2616#section-5.3).

artifacts/CAMARA_common.yaml
PedroDiez: Hi Kevin, 

Understood and OK to use kebab-case for the parameter property value. I checked with Jose, he was refering only to use UpperCamelCase for the schema name, just for aligment. Usually code generators also works better with schemas named following Camelcase. So it would read (just removing "-", for the new schemas included): 

  **IPv4Address:**
   in: header
   name: IPv4-Address
   required: false
   schema:
    $ref: "#/components/schemas/DeviceIpv4Address"
    
    
WDYT?

PedroDiez:
To summarize the comments:

**- Comment 1**: It can be used "UpperCamelCase" as model for schema convention which does not preclude to name a parameter which is a header as kebab-case notation

```
  IPv4-Address:
   in: header
   name: IPv4-Address
   required: false
   schema:
    $ref: "#/components/schemas/DeviceIpv4Address"
```

**- Comment 2**: With regards to the schema model to refer to a "header". 
- It would be needed to align on the serialization mode. (probably explode=true)
- At least IPv6-Address "header" definition points to schema that is an object so this would need to define the serialziation mode

cc @Kevsy (we talked in 25th March Commonalities meeting about this PR)

rartych:
@PedroDiez @Kevsy I see headers for device identification are used in https://github.com/camaraproject/SimpleEdgeDiscovery/pull/4/files 
All defined headers are simple strings, so the serialization is not needed.

The `IPv4-Address` header is defined as follows:
```
        - name: IPv4-Address
          in: header
          required: false

          schema:
            $ref: "#/components/schemas/SingleIpv4Addr"
```
Do you think this approach can be used here? But   replacing `SingleIpv4Addr` with proposed  `SingleIpv4Address` ?
This can be also the move in the direction indicated in https://github.com/camaraproject/Commonalities/issues/171 .

rartych:
The changes will be discussed again within https://github.com/camaraproject/Commonalities/issues/507

PR #143, opened 2024-02-19
Update Design Guidelines with character set guidance

### LINKED ISSUE #112: Guidance on the inclusion of charsets in content-type response headers ("application/json" vs "application/json;charset=utf-8")
**Problem description**
I have had the opportunity to review multiple Operator test deployments. For the `content-type` response header, I have seen some Operators return `application/json` and others return `application/json;charset=utf-8`.

Is there guidance on the inclusion of charsets?

**Possible evolution**
A paragraph within Design Guidelines documenting the group view of charset inclusion.

trehman-gsma:
Explicitly states that char sets must be UTF-8 to address issue #112

#### What type of PR is this?

* documentation

Explicitly states the supported character set is only UTF-8 as per #112 

Fixes #112 

Explicitly states that UTF-8 is the mandatory character set due to [RFC 8259 (JSON Data Interchange Format)](https://datatracker.ietf.org/doc/html/rfc8259#section-8.1)

```
JSON text exchanged between systems that are not part of a closed
   ecosystem MUST be encoded using UTF-8 [[RFC3629](https://datatracker.ietf.org/doc/html/rfc3629)].

   Previous specifications of JSON have not required the use of UTF-8
   when transmitting JSON text.  However, the vast majority of JSON-
   based software implementations have chosen to use the UTF-8 encoding,
   to the extent that it is the only encoding that achieves
   interoperability.
```

### DIFF
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -22,6 +22,7 @@ This document captures guidelines for the API design in CAMARA project. These gu
     - [3.4 Path Parameters Use](#34-path-parameters-use)
     - [3.5 HTTP Headers Definition](#35-http-headers-definition)
     - [3.6 MIME Types](#36-mime-types)
+      - [3.6.1 Content-Type character set](#361-content-type-character-set)
   - [4. API Resource Definition](#4-api-resource-definition)
     - [4.1 URL Definition](#41-url-definition)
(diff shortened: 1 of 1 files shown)


rartych (approved):

PedroDiez (approved):
LGTM

Issue #127, opened 2024-01-30
Which error send back when issue with Device Identifier provided by the API Consumer?


bigludo7:
**Problem description**
A very small issue but looking for feedback. I guess we can have a common behavior.

In a lot of our API we use the device identifier schema to get the identifier of the line. That's fine, but which error to send back if the device identifier is not supported by the API Implementation? example, an implementation is only able to perform an API with a phone number or IP but not networkAccessIdentifier.

**Possible evolution**
Agreed to use for this cas error 501 with code DEVICE_IDENTIFIER_TYPE_NOT_MANAGED

**Alternative solution**
use 500 (with code DEVICE_IDENTIFIER_TYPE_NOT_MANAGED) ?
Let this at the hand of each implem?

**Additional context**

shilpa-padgaonkar:
Agree that we need a common behaviour for this. How about 501 with code DEVICE_IDENTIFIER_TYPE_NOT_SUPPORTED instead of DEVICE_IDENTIFIER_TYPE_NOT_MANAGED

pjhac:
Hello,

I was looking at the RFC 9110 definition of 501: "the server does not support the functionality required to fulfill the request. This is the appropriate response when the server does not recognize the request method and is not capable of supporting it for any resource".

In the present case, I think (but I may be wrong in my understanding, please correct me if that's the case) that the server is actually able to handle the request, but not some of the parameters. If the supported parameters are known by the calling party, then I believe it should be a 400 as the calling party is able to change the request input parameters. Technically, a 501 cannot be fixed by the calling party (the code is not available on the server side to handle the request), as per the note inside [https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501).

To make an analogy, it would be like calling a function inside a program with a parameter that is unknown to the function; this is a caller error, as the function cannot be blamed for not supporting a parameter defined by the caller.

What do you think?

pjhac:
Good catch, @uwerauschenbach, I am with you for 422.

The example in RFC 9110 expresses it this way: "for example, this status code can be sent if an XML request content contains well-formed (i.e., syntactically correct), but semantically erroneous XML instructions."

Having a syntactically correct JSON with an inappropriate parameter is a semantic error, so I am in favor of it.

bigludo7:
Thanks for the suggestion. I'm fine with the 422 as well.
[...]

PR #145, opened 2024-02-21
Added 501 Not Supported error response

### LINKED ISSUE #138: Proposal to add error response for unsupported method for phone number
**Problem description**
CAMARA APIs may not support all phone numbers for specific functions in any point in time

**Possible evolution**
Add a 501 status response with the following message "Service not supported for this phone number"

```
NotSupported:
   description: Not Supported
   content:
    application/json:
     schema:
      $ref: "#/components/schemas/ErrorInfo"
     example:
      status: 501
      code: NOT_SUPPORTED
      message: Service not supported for this phone number
```

**Alternative solution**
N.A

**Additional context**
CAMARA APIs may support mobile lines first and then landlines later, vice versa.
A MNO may support the CAMARA API method, but the MVNO (of the same MNO) does not support it.

mhfoo:
#### What type of PR is this?

* enhancement

Enhancement as per #138 Proposal to add 501 Not Supported error response

Fixes #138

#### Changelog input

```
release-note
+ Add 501 Not Supported error response

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -265,6 +265,16 @@ components:
       status: 501
       code: NOT_IMPLEMENTED
       message: This functionality is not implemented yet
+  NotSupported:
+   description: Not Supported
+   content:
+    application/json:
+     schema:
+      $ref: "#/components/schemas/ErrorInfo"
+     example:
+      status: 501
+      code: NOT_SUPPORTED
+      message: Service not supported for this phone number
   AuthenticationRequired:
    description: Authentication Required
    content:


mhfoo:
@rartych @shilpa-padgaonkar @patrice-conil @RubenBG7 @PedroDiez 

Kindly review this pull request for merging please.

artifacts/CAMARA_common.yaml
```diff
@@ -265,6 +265,16 @@ components:
       status: 501
       code: NOT_IMPLEMENTED
       message: This functionality is not implemented yet
+  NotSupported:
```
PedroDiez, line 268:
```suggestion
  PhoneNumberNotSupported:

```

artifacts/CAMARA_common.yaml
```diff
@@ -265,6 +265,16 @@ components:
...
       code: NOT_IMPLEMENTED
       message: This functionality is not implemented yet
+  NotSupported:
+   description: Not Supported
+   content:
+    application/json:
+     schema:
+      $ref: "#/components/schemas/ErrorInfo"
+     example:
+      status: 501
```
PedroDiez, line 275: To be aligned with Issue#127 output (so far 422)

artifacts/CAMARA_common.yaml
```diff
@@ -265,6 +265,16 @@ components:
...
+      code: NOT_SUPPORTED
```
PedroDiez, line 276:
```suggestion
      code: PHONE_NUMBER_NOT_SUPPORTED

```

artifacts/CAMARA_common.yaml
mhfoo: @PedroDiez 
Suggest not to be specific to "phoneNumber", so this error can be reused.

artifacts/CAMARA_common.yaml
mhfoo: @PedroDiez
Suggest not to be specific to "Phone_Number", so this error can be reused.

artifacts/CAMARA_common.yaml
mhfoo: @PedroDiez 
Understand there is a difference between 4xx and 5xx HTTP error codes.
> The 4xx codes are intended for cases in which the client has erred, and the 5xx codes for the cases in which the server is aware that the server has erred.

In this context, the client cannot "correct" the request as the server is not supporting this specific request.

hdamker: @mhfoo please see my comment https://github.com/camaraproject/Commonalities/issues/127#issuecomment-2094346214 regarding your same statement there. 422 with appropriate description is definitely more helpful for the client then 501.

eric-murray: I think 501 is appropriate for functionality that has not _yet_ been implemented, but will be (or, at least, could be) implemented at some point in the future, whereas a 4XX error should be returned for a request that will never make sense and will never be supported (such as requesting SIM Swap status for a landline).

However desirable it is that all implementations support all possible functionality from the start, I do not think that will be the case.

So I support introducing a 501 error, but we need to be careful how it is used.

rartych:
@mhfoo As we agreed on the last Commonalities meeting could you modify your proposal to:

```
  NotSupported:
   description: Not Supported
   content:
    application/json:
     schema:
      $ref: "#/components/schemas/ErrorInfo"
     example:
      status:  422
      code: NOT_SUPPORTED
      message: Service not supported for this device
```

Further discussion is possible, but for now we have consensus to use 422.

mhfoo:
> @mhfoo As we agreed on the last Commonalities meeting could you modify your proposal to:
> 
> ```
>   NotSupported:
>    description: Not Supported
>    content:
>     application/json:
>      schema:
>       $ref: "#/components/schemas/ErrorInfo"
>      example:
>       status:  422
>       code: NOT_SUPPORTED
>       message: Service not supported for this device
> ```
> 
> Further discussion is possible, but for now we have consensus to use 422.

@rartych 

In last Commonalities meeting, my statement are as follows:

I do not want my PR #145 to block this coming meta-release (Fall24) of Commonalities. Please continue this PR #145 discussion for the next release (Spring25).

rartych:
@mhfoo  Thank you for clarification, I got it wrong.
Let's postpone this PR as you suggest.

hdamker:
@rartych @shilpa-padgaonkar propose to "convert to draft" if this is not an active review

hdamker:
There is now #374 as part of release r2.2, which clarifies the (restricted) use of 501 in line with the TSC decision:

> - Special Consideration for error 501 NOT IMPLEMENTED to indicate optional endpoint:
     - The use of optional endpoints is discouraged in order to have aligned implementations
     - Only for exceptions where an optional endpoint can not be avoided and defining it in separate, atomic API is not possible - error status 501 should be documented as a valid response

For the use which was intended in this draft PR there are other error responses defined within the API guidelines.
@rartych I propose to close this PR.

mhfoo:
Agree to close this PR, without merging.

PR #169, opened 2024-03-22
[Bugfix]: Linting rules problem with Traffic Influence API #161

### LINKED ISSUE #161: Linting rules problem with Traffic Influence API
**Problem description**
Using Spectral over the TI API a non correct error is returned.

https://github.com/camaraproject/EdgeCloud/issues/191#issuecomment-1956675044

**Expected action**
Correction of: https://github.com/camaraproject/Commonalities/blob/main/artifacts/linting_rules/.spectral.yml. 
Section starting on raw 103: "camara-http-methods:"

**Additional context**
According to my understanding, the tool doesn't recognize some methods as GET, PATCH and DELETE. Even if they actually are GET, PATCH and DELETE. I suppose the reason is that in the TI API, we defined "parameters" as child of "paths" rather than child of the "methods". Our implementation is correct according to the OpenAPI Specifications (https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#path-item-object). We have used the "parameters" description as child of "path". With our solution we avoid duplicating the parameter definition for the different methods (Get, Patch and Delete). With the suggested approach we should define the same parameter three times in the code with the risk of code misalignment.

Reference in the TI API: raw 2012
https://github.com/camaraproject/EdgeCloud/blob/32c7ca4faf54fc14fa510c1212f34a0f43a42bfd/code/API_definitions/Traffic%20Influence/Traffic_Influence.yaml

VijayKesharwani:
#### What type of PR is this?

* bug

Fixing camara-http-method rule to include parameters as valid identifier under path.

[
](https://github.com/camaraproject/Commonalities/issues/161)
Fixes #161 

Hello Team, 
As per the issue #161, this PR solves the problem of giving error on using 'parameters' under path for traffic Influence API yaml file related to camara-http-method linting rule.

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/artifacts/linting_rules/.spectral.yml
+++ b/artifacts/linting_rules/.spectral.yml
@@ -1,6 +1,6 @@
 # CAMARA Project - linting ruleset - documentation avaialable here:
 # https://github.com/camaraproject/Commonalities/blob/main/documentation/Linting-rules.md
-# 31.01.2024 - initial version
+# 19.03.2024 - camara-http-method rule corrected
 
 extends: "spectral:oas"
 functions:
--- a/artifacts/linting_rules/.spectral.yml
+++ b/artifacts/linting_rules/.spectral.yml
@@ -1,6 +1,9 @@
 # CAMARA Project - linting ruleset - documentation avaialable here:
 # https://github.com/camaraproject/Commonalities/blob/main/documentation/Linting-rules.md
-# 19.03.2024 - camara-http-method rule corrected
+# Changelog:
+# - 31.01.2024: Initial version
+# - 19.03.2024: Corrected camara-http-methods rule
+
 
 extends: "spectral:oas"
 functions:
--- a/artifacts/linting_rules/.spectral.yml
+++ b/artifacts/linting_rules/.spectral.yml
@@ -108,7 +108,7 @@ rules:
     then:
       function: pattern
       functionOptions:
-        match: "^(get|put|post|delete|patch|options)$"
+        match: "^(get|put|post|delete|patch|options|parameters)$"
     recommended: true  # Set to true/false to enable/disable this rule
 
   camara-get-no-request-body:


rartych:
In order to have simple versioning of linting configuration I would modify the line 3 to sth like:
`# 19.03.2024 - camara-http-method rule corrected`

artifacts/linting_rules/.spectral.yml
```diff
@@ -1,6 +1,6 @@
 # CAMARA Project - linting ruleset - documentation avaialable here:
 # https://github.com/camaraproject/Commonalities/blob/main/documentation/Linting-rules.md
-# 31.01.2024 - initial version
```
hdamker, line 3: If change comments are added with dates ... wouldn't it then make sense to keep the previous lines and create a kind of changelog by that?

rartych: Yes, it is possible way. I assumed that the changes can be tracked using file history in the Github, and files are marked with timestamp to easily verify if linting configuration in subproject is up to date with Commonalities artifacts.
Since it is the first correction of the linting ruleset we can start using the changelog style versioning if it doesn't clutter the files.

rartych (approved):
LGTM

hdamker (approved):

PedroDiez (approved):
LGTM

FabrizioMoggio:
Can this fix be approved?

rartych:
@FabrizioMoggio Is it possible for you to test it? 
I guess the fix works well,
If you approve it, it can be merged quickly.

FabrizioMoggio:
Tested in the TI API, it works.

rartych:
@FabrizioMoggio Thank you for your feedback.

PR #193, opened 2024-05-02
guideline_for_datacontenttype_cloudevent_concept_as_enum

PedroDiez:
#### What type of PR is this?

* correction
* enhancement/feature

#### What this PR does / why we need it:

By means of this PR, two topics are covered:
- Set `datacontenttype` cloudevent field value as enum "application\json", consistently with API design guidelines
- Take advantage of this PR to align success response to `204` only, aligned with PR #189

#### Which issue(s) this PR fixes:

Fixes #192 

#### Changelog input

```
 release-note

```

#### Additional documentation 

N/A

PR #189, opened 2024-04-18
Create subscription-notification-template.yaml

bigludo7:
Create subscription-notification-template.yaml aligned with CloudEvents subscription model.  Refer to https://github.com/camaraproject/Commonalities/issues/185

#### What type of PR is this?

Add one of the following kinds:

* documentation

#### What this PR does / why we need it:

Create subscription-notification-template.yaml aligned with CloudEvents subscription model.  

Refer to https://github.com/camaraproject/Commonalities/issues/185

#### Which issue(s) this PR fixes:

Fixes #185 

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

PR #194, opened 2024-05-02
x-correlator_guideline_format_convention

### LINKED ISSUE #191: X-Correlator or x-correlator
**Problem description**
Guidelines in Commonalities specifies "X-Correlator" but almost all API specs define already "x-correlator"

**Expected behavior**
Even if header name is case insensitive and both options are interoperable, we should define a common behaviour for coherence, aligned with Guidelines.

As de facto API specs are using "x-correlator", it would be less effort to reflect this in the Guidelines.

It would be convenient to set a new linting rule to check that the header is included in the requests and responses, following the agreed format.

**Alternative solution**
Adjust all APIs using "x-correlator"

PedroDiez:
#### What type of PR is this?

* correction

This PR covers the `x-correlator` naming guideline adoption.
Also given format to CAMARA_common.yaml (2-space indentation)

Documentation impacted (some doubts commented within PR):
- [API-design-guidelines.md](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md)
- [CAMARA_common.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml)
- [notification-as-cloud-event.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/notification-as-cloud-event.yaml)

Fixes #191 

Some notes:
1) To be merged after #193 
2) Should we take advantage of this PR and align error exceptions for POST /notifications? To align with #189. Commented within PR
3) Removed Exception 415
4) Refactor bearer format to bearerFormat: "{$request.body#/sinkCredential.credentialType}"

#### Changelog input

```
 release-note

```

#### Additional documentation 

N/A

### DIFF
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -339,3 +339,14 @@ components:
             message:
               The server refuses to accept the request because the payload format
               is in an unsupported format.
+    Gone:
+      description: Gone
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/ErrorInfo"
+          example:
+            status: 410
+            code: GONE
+            message:
+              Access to the target resource is no longer available.
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -7,7 +7,8 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.4.0
+  version: wip
+
 paths: {}
 components:
   headers:
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -1,326 +1,340 @@
 openapi: 3.0.3
 info:
- title: Camara common data types
- description: Common data types for Camara APIs
- contact:
-  email: sp-com@lists.camaraproject.org
- license:
-  name: Apache 2.0
-  url: https://www.apache.org/licenses/LICENSE-2.0.html
- version: 0.4.0
+  title: Camara common data types
+  description: Common data types for Camara APIs
+  contact:
+    email: sp-com@lists.camaraproject.org
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.4.0
 paths: {}
 components:
- schemas:
-  TimePeriod:
-   properties:
-    startDate:
-     type: string
-     format: date-time
-     description: An instant of time, starting of the TimePeriod.
-    endDate:
-     type: string
-     format: date-time
-     description: An instant of time, ending of the TimePeriod. If not included, then the period has no ending date.
-   required:
-    - startDate
-  ErrorInfo:
-   type: object
-   required:
-    - message
-    - status
-    - code
-   properties:
-    message:
-     type: string
-     description: A human readable description of what the event represent
-    status:
-     type: integer
-     description: HTTP response status code
-    code:
-     type: string
-     description: Friendly Code to describe the error
-  Device:
-   description: |
-    End-user equipment able to connect to a mobile network. Examples of devices include smartphones or IoT sensors/actuators.
-        The developer can choose to provide the below specified device identifiers:
-        * `ipv4Address`
-        * `ipv6Address`
-        * `phoneNumber`
-        * `networkAccessIdentifier`
-        NOTE: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device.
-   type: object
-   properties:
-    phoneNumber:
-     $ref: "#/components/schemas/PhoneNumber"
-    networkAccessIdentifier:
-     $ref: "#/components/schemas/NetworkAccessIdentifier"
-    ipv4Address:
-     $ref: "#/components/schemas/DeviceIpv4Addr"
-    ipv6Address:
-     $ref: "#/components/schemas/DeviceIpv6Address"
-   minProperties: 1
+  headers:
+    x-correlator:
+      description: Correlation id for the different services
+      schema:
+        type: string
+  parameters:
+    x-correlator:
+      name: x-correlator
+      in: header
+      description: Correlation id for the different services
+      schema:
+        type: string
+  schemas:
+    TimePeriod:
+      properties:
+        startDate:
+          type: string
+          format: date-time
+          description: An instant of time, starting of the TimePeriod.
+        endDate:
+          type: string
+          format: date-time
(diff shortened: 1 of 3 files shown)


PedroDiez:
cc @jlurien

artifacts/CAMARA_common.yaml
```diff
@@ -1,326 +1,340 @@
...
-  url: https://www.apache.org/licenses/LICENSE-2.0.html
- version: 0.4.0
+  title: Camara common data types
+  description: Common data types for Camara APIs
+  contact:
+    email: sp-com@lists.camaraproject.org
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+  version: 0.4.0
```
PedroDiez, line 10: Is it needed to move to 0.5.0?

artifacts/notification-as-cloud-event.yaml
```diff
@@ -89,23 +89,23 @@ paths:
...
         200:
           description: Ok
           headers:
-            X-Correlator:
-              $ref: "#/components/headers/X-Correlator"
+            x-correlator:
+              $ref: "#/components/headers/x-correlator"
           content:
             application/json:
               schema:
```
PedroDiez, line 111: Regarding Exceptions alignment with [PR#189](https://github.com/camaraproject/Commonalities/pull/189):
- `400`, `401`, `403`, `500`, `503` are clear
- `415`, `429` think makes sense to have in PR#189 as well
- `410` not sure about UC for 410. In case we keep it, we should align in PR#189 as well

cc @shilpa-padgaonkar, @patrice-conil, @bigludo7, @rartych, @jlurien

artifacts/CAMARA_common.yaml
bigludo7: I'm not sure about the management of the version for this document. 
My 2 cents: we should try to keep aligned this document version with the guideline version? so in this cas it should be 0.4.0-wip?

artifacts/notification-as-cloud-event.yaml
bigludo7: Ok for me to add 415 & 429 in PR#189 event if I'm bit dubious about the UC. I will do.

For 410 I'm more in favor to not add it but will follow your guidance.

artifacts/CAMARA_common.yaml
shilpa-padgaonkar: @bigludo7 Do you mean here the repo-release version? I could not find any version specified in the guidelines doc.

artifacts/notification-as-cloud-event.yaml
PedroDiez: My understanding:
415 - would be the case when API Consumer (Server Side in this flow) does not support "application/json" -> This in principle should never happen because is a guideline of CloudEvent CAMARA proifle
429 - would be the case when API Consumer (Server Side in this flow) receiving many requests over its threshold (could happen)

@patrice-conil Do you have context abiut why 410, 415, 429 were included in the initial proposal? https://github.com/camaraproject/Commonalities/pull/43/commits/339a8e2410dac6be4f6e533673ca03a27447edd4

PedroDiez: 410 - The unique case I think it may apply is the case of that callback no longer available. I mean the Server side decided to switch-off that endpoint. This would be differnet to a 503, because in 503 endpoint exists but is not available temporarily (regular exception for maintenance tasks)

bigludo7: So following your explanation @PedroDiez  I understand that we should have 429 & 410 and drop 415.

artifacts/CAMARA_common.yaml
bigludo7: I was thinking of the version in this document but I've probably mixing myself with the branch version :)

shilpa-padgaonkar: :) Then, we could simply have wip? 
(common.yaml version was already 0.4 in comms release 0.3, so we would need to increment the version for the upcoming rel(0.4) to 0.5)

shilpa-padgaonkar, line 10:
```suggestion
  version: wip

```

bigludo7: Yes just wip will be good enough !

artifacts/notification-as-cloud-event.yaml
patrice-conil: Hi @PedroDiez,
My understanding:
410 was introduced to handle cases where the callback URL is no longer available.
415 is what the server side should respond if Content-type or Accept are not "application/json" (should never happen) .
429 could be used to handle congestion on the callback endpoint.

PedroDiez: Hi @patrice-conil, thanks for the clarification! I think we need to align with #189 (POST callback) cc @bigludo7 

My view:
- Keep 410 for the case the endpoint is no longer available (Not 503 case because that one is temporal). Benefit is to have an indication for Telco Operation teams to enable operations procedure to the Subscriber to say "Please be aware there is a flow no longer working, you will need to generate a new subscription with a new callback or in the case of implicit flow to indicate new callback"

- Would take out 415 - Should never happen because developers MUST adopt CAMARA cloudevent profile (if you think/prefer to have for safety/operation reasons, i.e. to advise an App developer that need to support application/json as a warning to adapt its side to be able to interoperate it is fine)

- Keep 429

WDYT?

PedroDiez: Done

PedroDiez:
ready for review
cc @shilpa-padgaonkar @rartych @patrice-conil @bigludo7

I have updated bearer format to: "{$request.body#/sinkCredential.credentialType}"

As https://swagger.io/docs/specification/authentication/bearer-authentication/ indicates, is an optional property and a hint for the client. ( # optional, arbitrary value for documentation purposes), so semantically speaking accordingly to CloudEvents model is the concept of credentialType (to me makes more sense but to have aligment/consensus on that)

If you think it is better to indicate: "{$request.body#/sinkCredential.accessToken}" just because we only allow ACCESSTOKEN so far and the format is "accessToken" (i.e. a string) it is also fine to me. 

Just to be transparent about this.

shilpa-padgaonkar (approved):
/LGTM

shilpa-padgaonkar:
@patrice-conil or @bigludo7 : Could you kindly review and approve the PR if you find no issues so that we can proceed to merge?

patrice-conil (approved):
LGTM

PedroDiez:
If no other comments I will be merging on Monday next week

bigludo7 (approved):
/LGTM

PR #227, opened 2024-06-07
Small component improvements & typo fixes

### LINKED ISSUE #226: [Subscription-CloudEvent-template]: fix typos and component improvements
**Problem description**
The template has some typos and unused components, e.g. no references inside of the `Subscription`-component for the property `id` to the component `SubscriptionId`

maxl2287:
#### What type of PR is this?

* bug
* correction

This PR: 
- fixes typos
- references inside of the `Subscription`-component for the property `id` to the component `SubscriptionId`

Fixes #226

### DIFF
--- a/artifacts/camara-cloudevents/event-subscription-template.yaml
+++ b/artifacts/camara-cloudevents/event-subscription-template.yaml
@@ -6,7 +6,7 @@ info:
 
     Note on ``security`` - ``openId`` scope: The value in this yaml `api-name:event-type:grant-level` must be replaced accordingly to the format defined in the guideline document.
 
-  termsOfService: http://swagger.io/terms/
+  termsOfService: https://swagger.io/terms/
   contact:
     email: project-email@sample.com
   license:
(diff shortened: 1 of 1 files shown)


artifacts/camara-cloudevents/event-subscription-template.yaml
```diff
@@ -447,20 +450,98 @@ components:
         - refreshToken
         - refreshTokenEndpoint
 
-    SubscriptionDetail:
-      description: The detail of the requested event subscription
-      type: object
-
     CreateSubscriptionDetail:
-      description: The detail of the requested event subscription
+      description: |
```
PedroDiez, line 454: Think we should not add any specific model within this object and left empty, every project will customize the content of this object based on its scope.

maxl2287:
> I think it is a matter of swagger editor.
> 
> With redocly it renderizes OK:https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/camaraproject/Commonalities/b14810ac4c2183e27ead588fa44fd50b9a353a9a/artifacts/camara-cloudevents/event-subscription-template.yaml&nocors
> 
> If we think on adding it:
> 
> * Fine to fix some typos (e .g. "SUBSCRIPTION_MISMATCH") and some editorial enhancements
> * Fine to adding example in request (Line 50 and example from line 1099. I think for response it would be needed something analogous as it is also not shown in swagger editor (and we can check it works for both swagger and redocly)
>   image
> 
> What i think we should not do is to define anything for "CreateSubscriptionDetail" object and remain empty:
> 
> * The aim of this template is to be a baseline template for any project
> * "device" object is applicable only for some projects. Every project will define specifically the "CreateSubscriptionDetail" content based on its scope

@PedroDiez Yes you are right. It's just a problem on Swagger UI, which is not be able to handle discriminator examples.
Redocly can do that pretty well.

So I decided to change this PR to just do the small improvements, without the example adding (as this would remove the examples with redocly).

PedroDiez:
Thanks for the explanation @maxl2287, ok with the improvements

PedroDiez (approved):
LGTM

rartych (approved):
LGTM

patrice-conil (approved):
LGTM

PR #233, opened 2024-06-11
Proposal to simplify device object and improve interoperability

### LINKED ISSUE #171: Revise the device object definition to simplify it
**Problem description**

In the current existing device object schema, the developer can choose to provide any of the following identifiers: 
 
* ipv4Address, with 2 flavours 
* ipv6Address 
* phoneNumber 
* networkAccessIdentifier 
 
By assuming that each implementation will support a subset of identifier types, we are going against interoperability. A developer will have to code their application to collect and send every available value to maximize/ensure success. This is not a good practice and, in our view, it is not the way to favour interoperability. 
 
At Telefónica, we believe that the current device object schema makes  interoperability harder if not all identifiers in the specification are fully supported, or if we don't agree on certain rules about a common subset that should be fully supported. Agreeing on error codes to overcome this is not the long term solution.

**Possible evolution**

Review and simplify the device object definition with the following considerations: 
 
* Depending on the functionality provided by the CAMARA API, some API subprojects may still define other specific identifiers that differs from the common device object definition proposed here. Not all APIs necessarily have to refer to a user device, e.g. Carrier Billing API only defines a phone number as a way to identify the mobile account to be billed, Know Your Costumer only defines a phone number as a way to identify the associated account data, Home Devices QoD API defines public IP v4 address as a way to identify the user home network, etc. 
 
* Minimize the number of required identifiers **as a common subset to be supported by all implementations**. Telefónica suggests using the phone number, and the IMSI: 
  * Phone number -> We may agree that the phone number is generally the most relevant identifier and it is suitable in most scenarios. 
  * IMSI -> to cover mobile scenarios with devices that do not have a phone number associated, e.g. M2M SIM cards. Other options such as ICCID or IMEI could be discussed as an alternative identifier to IMSI. We need to choose one of them that best covers the required scenarios. 
 
* Provide a mechanism for optionally specifying that the device to be considered in the API request is the one associated with the access token used to invoke the API (making potentially device object not necessary in the request). This makes sense for three-legged access tokens.  
  * This avoids having to enter multiple identifiers and validate them against each other and against the identifier associated with the access token. Developers would be able to rely directly on the authenticated identifier associated with the access token and simplify the required validations on the operator side when using 3-legged access tokens. 
  * In case the access token is a 3-legged token, but the API client still provide a device identifier, the API should validate that the provided device identifier matches the one associated with the access token. If they do not match, the API must return an error. 
  * 2-legged access tokens would be not allowed to use this mechanism, as they do not have an associated authenticated identifier. And in this case, the API client must provide a device identifier in the request.

**Alternative solution**

* There is an ongoing discussion on https://github.com/camaraproject/Commonalities/issues/118 about specifying the identifiers as HTTP headers. We are also open to discussing this option as a potential alternative to providing the device identifiers in the request body. It might also allow us to use GET operations instead of POST operations with the request body to protect the device information (as it is not recommended to use device information as query parameter in GET operations). **However, sending identifiers as HTTP headers does not solve the underlying problem of simplifying the device object as raised in this issue and it should still be considered even if HTTP header are used instead.** 

**Additional context**

Regarding IPv4 / IPv6 identifiers, if this information is available, it means that there is a mobile/fixed CSP connection, so the best thing to do is to use Frontend Flow in any case, as described [here](https://github.com/camaraproject/IdentityAndConsentManagement/blob/main/documentation/CAMARA-API-access-and-user-consent.md#authorization-code-flow-frontend-flow). Then the API client does not need to retrieve the IP from the device as it is automatically bound to the access token. This way the limitations on the operator side to support IPv4 and/or IPv6 are hidden to the developer and the operator can use the supported mechanism as part of the network based authentication applied in the auth code flow. It is assumed that if a developer can't execute the front-end flow, they will need to have other data to refer to the device: phone number or IMSI. The requirement to explicitly use the IP address, such as in the [Simple Edge Discovery](https://github.com/camaraproject/EdgeCloud/blob/main/code/API_definitions/Discovery/simple_edge_discovery.yaml) API, could be limited to those APIs that have that specific requirement. 
 
And as for the Network Access Identifier, originally defined for use with the [GPSI External Identifier](https://www.tech-invite.com/3m23/toc/tinv-3gpp-23-501_zg.html#e-5-9-8), it may not be very developer friendly, and we suggest not including it as a device identifier, as it does not seem reasonable to mandate its support for all implementations unless there is a specific requirement to do so.

jpengar:
#### What type of PR is this?

* enhancement/feature
* documentation

PR incorporating the existing DRAFT proposal to simplify the device object and improve interoperability as per issue #171. 

>The solution proposed to be covered in the scope of this meta release was:
>
>1. Apply the mechanism to rely on the access_token (not providing the device object in the API request) for 3-legged access scenarios. This would fix interoperability in all these scenarios, which is a huge improvement. We can define the device identifier as optional in the API specification and, if not provided, simply refer to access_token. The developer would simply not provide any further device information, which is simpler. The operator does not need to check that the device identifier actually matches the access_token provided and could simply rely on the access_token itself.
>2. Validate the current situation among operators regarding the support and use of the networkAccessIdentifier in order to agree on the removal of this identifier for this meta-release.
>At least these two points would be a big improvement over the current situation. To cover all 3-legged scenarios, and in other cases where it is necessary to explicitly provide the device object in the request (such as 2-legged access with client credentials), we would have a smaller set of identifiers that would mitigate the problem a bit. So far, the APIs we are discussing (Device Location family, Device Status, Quality On demand) are mostly consumed with 3-legged access_tokens.

Reference: https://github.com/camaraproject/Commonalities/issues/171#issuecomment-2129218755

Fixes #171

**DISCLAIMER: The proposal is a minimum proposal that does not aim to solve all existing problems, and it is recognised that it does not cover all possible future use cases. However, in the context of the next meta-release, the two proposed measures could already be of great benefit.**

- Please find relevant context in #171
- **As agreed at the 10 June WG meeting, the DRAFT proposal has been included as an appendix to the API Design Guidelines for the time being. But we would need to decide what is the best place to document it. It could be as an appendix, in another specific document in the Commonalities, or even document it in ICM if the mechanism for relying on the access token is considered more appropriate for the ICM (e.g. in "CAMARA APIs access and user consent management" doc).**

#### Changelog input

```
- Device object simplification
- Mechanism to rely on access token to obtain device information
```

#### Additional documentation 

[2024-05-23 - Commonalities Ad-Hoc Meeting - Device Object review](https://wiki.camaraproject.org/display/CAM/2024-05-23+-+Commonalities+Ad-Hoc+Meeting+-+Device+Object+review)

### DIFF
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -60,7 +60,7 @@ components:
         * `phoneNumber`
         * `networkAccessIdentifier`
         NOTE1: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device.
-        NOTE2: for the Commonalities meta-release v0.4, we are enforcing that the networkAccessIdentifier is only part of the schema for future-proofing, and CAMARA does not currently allow its use. After the meta-release work is done and the relevant issues are resolved, its use will need to be explicitly documented in the guidelines.
+        NOTE2: for the Commonalities release v0.4, we are enforcing that the networkAccessIdentifier is only part of the schema for future-proofing, and CAMARA does not currently allow its use. After the CAMARA meta-release work is concluded and the relevant issues are resolved, its use will need to be explicitly documented in the guidelines.
    type: object
    properties:
     phoneNumber:
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -59,7 +59,8 @@ components:
         * `ipv6Address`
         * `phoneNumber`
         * `networkAccessIdentifier`
-        NOTE: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device.
+        NOTE1: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device.
+        NOTE2: for the Commonalities meta-release v0.4, we are enforcing that the networkAccessIdentifier is only part of the schema for future-proofing, and CAMARA does not currently allow its use. After the meta-release work is done and the relevant issues are resolved, its use will need to be explicitly documented in the guidelines.
    type: object
    properties:
     phoneNumber:
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -59,14 +59,11 @@ components:
         * `ipv4Address`
         * `ipv6Address`
         * `phoneNumber`
-        * `networkAccessIdentifier`
         NOTE: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device.
    type: object
    properties:
     phoneNumber:
      $ref: "#/components/schemas/PhoneNumber"
-    networkAccessIdentifier:
-     $ref: "#/components/schemas/NetworkAccessIdentifier"
     ipv4Address:
      $ref: "#/components/schemas/DeviceIpv4Addr"
     ipv6Address:
@@ -78,11 +75,6 @@ components:
    type: string
    pattern: '^\+[1-9][0-9]{4,14}$'
    example: "+123456789"
-
-  NetworkAccessIdentifier:
-   description: A public identifier addressing a subscription in a mobile network. In 3GPP terminology, it corresponds to the GPSI formatted with the External Identifier ({Local Identifier}@{Domain Identifier}). Unlike the telephone number, the network access identifier is not subjected to portability ruling in force, and is individually managed by each operator.
-   type: string
-   example: "123456789@domain.com"
 
   DeviceIpv4Addr:
    type: object
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -58,11 +58,14 @@ components:
         * `ipv4Address`
         * `ipv6Address`
         * `phoneNumber`
+        * `networkAccessIdentifier`
         NOTE: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device.
    type: object
    properties:
     phoneNumber:
      $ref: "#/components/schemas/PhoneNumber"
+    networkAccessIdentifier:
+     $ref: "#/components/schemas/NetworkAccessIdentifier"
     ipv4Address:
      $ref: "#/components/schemas/DeviceIpv4Addr"
     ipv6Address:
@@ -74,6 +77,11 @@ components:
    type: string
    pattern: '^\+[1-9][0-9]{4,14}$'
    example: "+123456789"
+
+  NetworkAccessIdentifier:
+   description: A public identifier addressing a subscription in a mobile network. In 3GPP terminology, it corresponds to the GPSI formatted with the External Identifier ({Local Identifier}@{Domain Identifier}). Unlike the telephone number, the network access identifier is not subjected to portability ruling in force, and is individually managed by each operator.
+   type: string
+   example: "123456789@domain.com"
 
   DeviceIpv4Addr:
    type: object
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -1822,7 +1822,7 @@ response:
 
 The documentation template below must be used as part of the API documentation in `info.description` property in the CAMARA API specs which use the `device`object defined in [CAMARA_common.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml) artifact. This template provides guidance on how to handle device information in API requests **when using 3-legged access tokens and the device can be uniquely identified by the token**.
 
-Note: With the current 3-legged authorization flows, only a single end user can be associated with the access token. For the OIDC authorization code flow, only a single device can call the `/authorize` endpoint and obtain the code. And for CIBA, `login_hint` is currently limited to a single phone number or IP address (that can optionally include a port).
+Note: With the current 3-legged authorization flows used by CAMARA, only a single end user can be associated with the access token. For the OIDC authorization code flow, only a single device can call the `/authorize` endpoint and obtain the code. And for CIBA, `login_hint` is currently limited to a single phone number or IP address (which can optionally include a port).
 
 ```md
 # Identifying a user device from the access token
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -50,7 +50,10 @@ This document captures guidelines for the API design in CAMARA project. These gu
         - [Polymorphism](#polymorphism)
     - [11.6 Security definition](#116-security-definition)
       - [11.6.1 Scope naming](#1161-scope-naming)
-        - [Examples](#examples)
+      - [Regarding scope naming for APIs which do not deal with explicit subscriptions, the guidelines are:](#regarding-scope-naming-for-apis-which-do-not-deal-with-explicit-subscriptions-the-guidelines-are)
+          - [Examples](#examples)
+      - [Regarding scope naming for APIs which deal with explicit subscriptions, the guidelines propose some changes as compared to the above format and this is described below:](#regarding-scope-naming-for-apis-which-deal-with-explicit-subscriptions-the-guidelines-propose-some-changes-as-compared-to-the-above-format-and-this-is-described-below)
+      - [API-level scopes (sometimes referred to as wildcard scopes in CAMARA)](#api-level-scopes-sometimes-referred-to-as-wildcard-scopes-in-camara)
   - [12. Subscription, Notification \& Event](#12-subscription-notification--event)
     - [12.1 Subscription](#121-subscription)
       - [Instance-based (implicit) subscription](#instance-based-implicit-subscription)
@@ -66,6 +69,7 @@ This document captures guidelines for the API design in CAMARA project. These gu
       - [Security Considerations](#security-considerations)
       - [Abuse Protection](#abuse-protection)
       - [Notification examples](#notification-examples)
+  - [Appendix A: `info.description` template for `device` identification from access token](#appendix-a-infodescription-template-for-device-identification-from-access-token)
 
 
 ## Common Vocabulary and Acronyms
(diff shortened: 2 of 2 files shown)


documentation/API-design-guidelines.md
```diff
@@ -50,7 +50,10 @@ This document captures guidelines for the API design in CAMARA project. These gu
         - [Polymorphism](#polymorphism)
     - [11.6 Security definition](#116-security-definition)
       - [11.6.1 Scope naming](#1161-scope-naming)
-        - [Examples](#examples)
+      - [Regarding scope naming for APIs which do not deal with explicit subscriptions, the guidelines are:](#regarding-scope-naming-for-apis-which-do-not-deal-with-explicit-subscriptions-the-guidelines-are)
```
PedroDiez, line 53:
```suggestion
        - [Regarding scope naming for APIs which do not deal with explicit subscriptions, the guidelines are:](#regarding-scope-naming-for-apis-which-do-not-deal-with-explicit-subscriptions-the-guidelines-are)
```

documentation/API-design-guidelines.md
```diff
@@ -50,7 +50,10 @@ This document captures guidelines for the API design in CAMARA project. These gu
...
+          - [Examples](#examples)
```
PedroDiez, line 54:
```suggestion
              - [Examples](#examples)
```

documentation/API-design-guidelines.md
PedroDiez: just add indentation to have analogous Look& Feell like subsections 11.5.1

documentation/API-design-guidelines.md
```diff
@@ -50,7 +50,10 @@ This document captures guidelines for the API design in CAMARA project. These gu
...
+      - [Regarding scope naming for APIs which deal with explicit subscriptions, the guidelines propose some changes as compared to the above format and this is described below:](#regarding-scope-naming-for-apis-which-deal-with-explicit-subscriptions-the-guidelines-propose-some-changes-as-compared-to-the-above-format-and-this-is-described-below)
```
PedroDiez, line 55:
```suggestion
        - [Regarding scope naming for APIs which deal with explicit subscriptions, the guidelines propose some changes as compared to the above format and this is described below:](#regarding-scope-naming-for-apis-which-deal-with-explicit-subscriptions-the-guidelines-propose-some-changes-as-compared-to-the-above-format-and-this-is-described-below)
```

documentation/API-design-guidelines.md
```diff
@@ -50,7 +50,10 @@ This document captures guidelines for the API design in CAMARA project. These gu
...
+      - [API-level scopes (sometimes referred to as wildcard scopes in CAMARA)](#api-level-scopes-sometimes-referred-to-as-wildcard-scopes-in-camara)
```
PedroDiez, line 56:
```suggestion
        - [API-level scopes (sometimes referred to as wildcard scopes in CAMARA)](#api-level-scopes-sometimes-referred-to-as-wildcard-scopes-in-camara)
```

PedroDiez (approved):
Minor indentation comment

In my view proposal and information allocation is fine

LGTM in advance

PedroDiez (dismissed):
LGTM

PedroDiez (dismissed):
LGTM

artifacts/CAMARA_common.yaml
```diff
@@ -59,14 +59,11 @@ components:
         * `ipv4Address`
         * `ipv6Address`
         * `phoneNumber`
-        * `networkAccessIdentifier`
```
Elisabeth-Ericsson, line 62: We do not think that it is reasonable to remove the networkAccessIdentifier, even though it is not widely used yet.
We actually would like to add a further option to allow identification via the operator token.
For security reasons we see a mandatory need to allow verification of the data subject named in the access token against the target resource, which is input to the api call. We need to keep all option open, especially since we are discussing (in issue #145 in ICM group requesting to support the operator token).

documentation/API-design-guidelines.md
```diff
@@ -1704,3 +1713,40 @@ response:
...
+# Identifying a user device from the access token
+
+This specification defines the `device` object field as optional in API requests, specifically in cases where the API is accessed using a 3-legged access token and the device can be uniquelly identified by the token. This approach simplifies API usage for API consumers by relying on the device information associated with the access token used to invoke the API. This mechanism reduces the need for multiple identifiers and extensive validation.
+
+## Handling of device information:
+
+### Optional device object for 3-legged tokens:
+
+- When using a 3-legged access token, the device associated with the access token must be considered as the device for the API request. This means that the device object is not required in the request, and if included it must identify the same device, therefore **it is recommended NOT to include it in these scenarios** to simplify the API usage and avoid additional validations.
+
```
Elisabeth-Ericsson, line 1732: We (E///) think that we rather should recommend to have the additional validation in place and compare the target device in the access token against the target device named on the API input. This gives additional security and allows for a better reporting in line with GDPR.

jpengar: I don't see how comparing the device identifier in the API request to the target device in the access token provides any additional security. And if we take the access token information as the source of truth, if we can rely directly on the access token, we would avoid additional validations that add complexity and new potential points of failure. In addition to improving interoperability as explained in https://github.com/camaraproject/Commonalities/issues/171#issue-2224959588

artifacts/CAMARA_common.yaml
jpengar: **We are focusing on a proposal in the scope of the next meta release**. The proposal to remove the networkAccessIdentifier is precisely because the feedback so far has been favourable to doing so, and there is no explicit feedback from any WG participant that they are currently using it. Of course, it could be revisited for future releases if the WG needs it, and it could be re-evaluated whether or not to include it again.

The operatorToken discussed in [ICM [#145](https://github.com/camaraproject/Commonalities/pull/145)](https://github.com/camaraproject/IdentityAndConsentManagement/issues/145) is just an open discussion for now (in ICM backlog), so I would suggest not even considering it for this meta release. In fact, operatorToken might fit as networkAccessIdentifier from our point of view, but it should be discussed for a release when there is a clear position and solution defined in CAMARA for operatorToken.

shilpa-padgaonkar: Can we try to find a middle way here? I would propose that we could write in the guidelines that the networkAccessIdentifier is only part of the schema to be able to be future-proof and Camara currently does not permit its use. 
Post meta-release work, and after the resolution of the relevant issues in ICM, its use will have to be explicitly documented in ICM and in the guidelines. 

This will be similar to what we have done for subscriptions where even though event type is an array (for being future-proof), we are allowing for the meta release the use of just single event type and after the meta-release we will target usecases which need subscriptions to support multiple event types.

WDYT?

jpengar: If it works and helps us reach an agreement, that would be fine with me. But I think it would be better to just have an artifact definition of what needs to be supported and allowed to be used. And then just evolve it in the next releases.

shilpa-padgaonkar: @Elisabeth-Ericsson : Could you please confirm if this suggested middle way would be ok for you as well?

Elisabeth-Ericsson: I'm fine with the middle way.

jpengar: > This will be similar to what we have done for subscriptions where even though event type is an array (for being future-proof), we are allowing for the meta release the use of just single event type and after the meta-release we will target usecases which need subscriptions to support multiple event types.

But in this case, the WG is also changing the schema in https://github.com/camaraproject/Commonalities/pull/236 to add minItems and maxItems set to "1". So with this approach, we would actually have a schema according to what will be supported in the next meta-release. Doing the analogy with our case in this PR, it would be removing NAI from the device object definition, so the artifact definition of what needs to be supported and allowed to be used.

I could try to suggest some changes in the PR to address this "middle way" in any case...

jpengar: Middle way agreement applied in https://github.com/camaraproject/Commonalities/pull/233/commits/4ae03bb1d76be80243badfdce988a8b254df0b76
Please have a look. shilpa-padgaonkar @Elisabeth-Ericsson and suggest any necessary changes to the wording I use.

jpengar: If an API client only provides the networkAccessIdentifier, we can return a 422 error, and if the networkAccessIdentifier is provided along with other allowed identifiers, it should be ignored.

Should we also add this to the NOTE included in the common.yaml file?

artifacts/CAMARA_common.yaml
```diff
@@ -59,7 +59,8 @@ components:
         * `ipv6Address`
         * `phoneNumber`
         * `networkAccessIdentifier`
-        NOTE: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device.
+        NOTE1: the MNO might support only a subset of these options. The API invoker can provide multiple identifiers to be compatible across different MNOs. In this case the identifiers MUST belong to the same device.
+        NOTE2: for the Commonalities meta-release v0.4, we are enforcing that the networkAccessIdentifier is only part of the schema for future-proofing, and CAMARA does not currently allow its use. After the meta-release work is done and the relevant issues are resolved, its use will need to be explicitly documented in the guidelines.
```
shilpa-padgaonkar, line 63:
```suggestion
        NOTE2: for the Commonalities release v0.4, we are enforcing that the networkAccessIdentifier is only part of the schema for future-proofing, and CAMARA does not currently allow its use. After the Camara meta-release work is concluded and the relevant issues are resolved, its use will need to be explicitly documented in the guidelines.

```

documentation/API-design-guidelines.md
```diff
@@ -1800,3 +1817,41 @@ response:
...
+### Error handling for unidentifiable devices:
+
+- If the `device` object is not included in the request and the device information cannot be derived from the 3-legged access token, the server will return a 422 `UNIDENTIFIABLE_DEVICE` error.
+
+### Restrictions for tokens without an associated authenticated identifier:
+
+- For scenarios which do not have a single device identifier associated to the token during the authentication flow, e.g. 2-legged access tokens, the `device` object MUST be provided in the API request. This ensures that the device identification is explicit and valid for each API call made with these tokens.
+
+By following these guidelines, API consumers can use the authenticated device identifier associated with 3-legged access tokens, simplifying implementation and validation. This mechanism ensures that device identification is handled efficiently and securely, and appropriately accommodates different token types.
+```
```
hdamker, line 1853: I suppose that the sentence in line 1852 does not belong any more to the template for the API specification but is an explanation of the guideline and the template itself. If this is right, then the end of the code block has to be already within line 1851.

````suggestion
```

By following these guidelines, API consumers can use the authenticated device identifier associated with 3-legged access tokens, simplifying implementation and validation. This mechanism ensures that device identification is handled efficiently and securely, and appropriately accommodates different token types.

````

File: documentation/API-design-guidelines.md [hunk 9]
```diff
@@ -1800,3 +1817,41 @@ response:
...
+
+Note: With the current 3-legged authorization flows used by CAMARA, only a single end user can be associated with the access token. For the OIDC authorization code flow, only a single device can call the `/authorize` endpoint and obtain the code. And for CIBA, `login_hint` is currently limited to a single phone number or IP address (which can optionally include a port).
+
+```md
+# Identifying a user device from the access token
+
+This specification defines the `device` object field as optional in API requests, specifically in cases where the API is accessed using a 3-legged access token and the device can be uniquelly identified by the token. This approach simplifies API usage for API consumers by relying on the device information associated with the access token used to invoke the API. This mechanism reduces the need for multiple identifiers and extensive validation.
```
hdamker, line 1830: If this text is meant as a template for the API specification, then the audience is primary the API consumer. For them the reduction of "extensive validation" is not relevant. And that the need for "multiple identifiers" is reduced is already included within the sentence before. Therefore I propose to omit the last sentence in this paragraph:

```suggestion
This specification defines the `device` object field as optional in API requests, specifically in cases where the API is accessed using a 3-legged access token and the device can be uniquelly identified by the token. This approach simplifies API usage for API consumers by relying on the device information associated with the access token used to invoke the API.
```

hdamker (changes requested):
Some suggestion for changes. Especially the template should not be strictly mandatory ("must") as we identified the potential need in QoD for cases where additional information from the device object might be needed. Even if we haven't identified them in detail yet, we want to have the option to describe them in our documentation.

Personal note: I'm not sure if all developers can decide if "the device can be uniquelly identified by the token". If unsure they might provide additional information within the device object, effectively creating the situation that the token is another parameter beside the ones within the device object. Some of them might be validated, others ignored (if not supported).

shilpa-padgaonkar (approved):
/LGTM

patrice-conil (approved):
LGTM

hdamker (approved):

PedroDiez (approved):
LGTM

artifacts/CAMARA_common.yaml
javierlozallu: This clarification is very useful, could you confirm if we can add this in our API subprojects?

rartych: > This clarification is very useful, could you confirm if we can add this in our API subprojects?

There is general guideline in API Design Guidelines section [6.2 Error Responses - Device Object](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#62-error-responses---device-object).
The clarification above is in line with it, so it can be added in subprojects.
If more detailed guideline is needed for networkAccessIdentifier it can be added as a fix for API Design Guidelines.

rartych (approved):

PR #314, opened 2024-10-10
Update Pull Request template with breaking change indication

rartych:
#### What type of PR is this?

* subproject management

Since some of CAMARA APIs are published as stable implementation of braking changes would require bumping up the main API version.
This should be indicated in the PR. And analyzed in its review.

#### Changelog input

```
Pull Request template with breaking change indication

```

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 #### What type of PR is this?
 
-Add one of the following kinds:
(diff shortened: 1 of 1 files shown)


eric-murray (approved):

PedroDiez (approved):
LGTM

jpengar:
It may be a good idea to incorporate this change into the ICM and other sub-projects.

Issue #284, opened 2024-08-15
Missing guideline for file names of API definition files - mixed use of snake_case and kebab-case


hdamker:
**Problem description**

There is currently no guidelines for file names of API definition files in CAMARA. 

The release candidates for Fall24 are all using the API name together with .yaml, but a mix of snake_case and kebab-case. For consistency in the meta-release and to be prepared for future automation with scripts this must be harmonized.

The API Design Guideline has already a definition of `api-name` in chapter [11.6.1 Scope Naming](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#1161-scope-naming):

> `api-name` is the API name specified as the base path, prior to the API version, in the servers[*].url property. For example, from `/location-verification/v0`, it would be `location-verification`

Within the path the api name must be in kebab-case. The proposal is therefore to use the convention `api-name.yaml` as a guideline for the API definition file names.

**Possible evolution**

- Define within the next version of Commonalities a guideline for the API Definitions file names
- Ensure that all Commonality artefacts are using the same convention to make their use easier (e.g. CAMARA_common.yaml vs notification-as-cloud-event.yaml in the same directory) (don't do it for 0.4.0 to avoid broken links).

The changes shouldn't be done in Commonalities 0.4.0 to avoid broken links, but a decision for the format should be taken, so that API Sub Project voluntary can do the change already for the Fall24 release, e.g. as part of their release PR for M4.

**Alternative solution**

**Additional context**

PedroDiez:
From TEF side we are agree with this issue and its resolution:
* To formally define it in the next version of commonalities
* Up to each WG to consider in advance for current MetaRelease in APIs

bigludo7:
From Orange perspective agree as well on the issue & resolution.

shilpa-padgaonkar:
@rartych Could you kindly create an issue in the subprojects identified by @hdamker above so that they can already change the file names. @bigludo7 has already created it for simswap.

PR #332, opened 2024-11-06
 Clarifications on api-name, filenames and servers object

### LINKED ISSUE #298: Further clarifying statement for api-name
**Problem description**

API Design Guidelines give a clear statement on api-name.
`api-name is the API name specified as the base path, prior to the API version, in the servers[*].url property. For example, from /location-verification/v0, it would be location-verification.`

However, since 'servers' field is an array of Server Object, there could be more than one `servers[*].url` property.  An API provider may use different strings in each of the Server Objects resolving to different `api-name`.

**Expected action**

Update API Design Guideline to ensure same 'api-name' resolution even if there are multiple server objects defined.

**Additional context**

rartych:
#### What type of PR is this?

* correction

Missing guideline for file names of API definition files, indirect definition of `api-name` variable.
This PR adds the definition of api-name to Common Vocabulary and Acronyms and clarifies the name of API definition files and servers object content 

Fixes #298 #284

- [ ] Yes
- [x] No

Do we need similar definition of 'api-version'?
JSON format is not currently used in any subproject, should the file extension .json be removed?

#### Changelog input

```
 Clarifications on api-name, filenames and servers object

```


rartych:
Closing due to problem with EasyCLA - fixed in the PR #333

PR #313, opened 2024-10-10
(WIP) Subscription & events for Commonalities 0.5.0

bigludo7:
#### What type of PR is this?

Add one of the following kinds:

* documentation

#### What this PR does / why we need it:

This PR covers update to subscriptions & events part for Commonalities 0.5.0

- [ ] https://github.com/camaraproject/Commonalities/issues/304
Resolution: Scope larger than subscription - proposal to manage it in other part of the document. 
- [ ] https://github.com/camaraproject/Commonalities/issues/303
Resolution: to be done
- [ ] https://github.com/camaraproject/Commonalities/issues/300
Resolution: To be done - seems we have an alignement to make device optional
- [ ] https://github.com/camaraproject/Commonalities/issues/299
Resolution: Discussion still open in the issue
- [ ] https://github.com/camaraproject/Commonalities/issues/295
Resolution: Discussion still open in the issue
- [ ] https://github.com/camaraproject/Commonalities/issues/291
Resolution: Waiting for https://github.com/camaraproject/Commonalities/issues/308 resolution
- [ ] https://github.com/camaraproject/Commonalities/issues/290
Resolution: Very simple one but need to be done once we split the document. Probably to be removed from this PR as we'll merge this one before the splitting (action @rartych )
- [x] https://github.com/camaraproject/Commonalities/issues/289
Resolution: covered in this PR
- [ ] https://github.com/camaraproject/Commonalities/issues/276
Resolution: 2 proposals - waiting for team feedback
- [x] https://github.com/camaraproject/Commonalities/issues/243
Resolution: covered in this PR

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for reviewers:

**Not ready for review**

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

PR #334, opened 2024-11-07
Resource access restriction

### LINKED ISSUE #304: Resource-based subscription – GET /subscriptions using a 3-legged access token
**Problem description**
When creating a resource-based subscription to information related to a device (such as location, roaming status etc) there is today in the POST /subscription subscriptionDetail a device object to indicate which device the subscription relates to.

In case a 3-legged access token is used it is assumed that “the device associated with the access token must be considered as the device for the API request. This means that the device object is not required in the request, and if included it must identify the same device.”

But how should the usage of a 3-legged access token be interpreted when used with GET /subscriptions, GET /subscriptions/{subscriptionId} or DELETE /subscriptions/{subscriptionId} in which the device is not part of the request parameters?

**Possible evolution**
Describe the expected behavior, in the API-design-guidelines Subscription chapter, when using a 3-legged access token when invoking

1. POST /subscriptions in cases when there is a device object in the subscriptionDetail
2. GET /subscriptions and GET /subscriptions/{subscriptionId}
3. DELETE /subscriptions/{subscriptionId}

**Alternative solution**
On one hand it would be natural to view device information in the access token as a filter. On the other hand it would be difficult for an application to get a list of all its subscriptions no matter which device it involves with less than having a separate 2-legged access token for that purpose.

**Additional context**

rartych:
#### What type of PR is this?

* enhancement/feature

Guidelines for access restrictions to resources created for given end user or API Customer, based on discussion in the issue #304 

Fixes #304 

- [x] Yes
- [ ] No

The issues started with question related to subscriptions.
Should this guidelines be repeated in event subscription guidelines or only referenced?

#### Changelog input

```
Guidelines for resource access restriction added in 11.7

```

### DIFF
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -1567,6 +1567,28 @@ makes scope name: `device-roaming-subscriptions:org.camaraproject.device-roaming
 The decision on the API-level scopes was made within the [Identity and Consent Management Working Group](https://github.com/camaraproject/IdentityAndConsentManagement) and is documented in the design guidelines to ensure the completeness of this document. 
 The scopes will always be those defined in the API Specs YAML files. Thus, a scope would only provide access to all endpoints and resources of an API if it is explicitly defined in the API Spec YAML file and agreed in the corresponding API subproject. 
 
+### 11.7 Resource access restriction 
+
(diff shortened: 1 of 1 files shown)


rartych:
The issue is covered in PR #313

PR #339, opened 2024-11-15
Add VERSION.yaml file

### LINKED ISSUE #307: Commonalities release versioning
**Problem description**
The current proposal for the "name" (i.e. version) of the next Commonalities release is "0.5.0". The issues with using semantic versioning for Commonalities documentation are:
- Using semantic versioning with major version 0 rather begs the question as to when the Commonalities documentation will become "stable". There are no agreed criteria that would result in a major version bump.
- The release is a set of documents and artifacts that may have their own separate release number. Using semantic versioning for both the release as a whole and documents within may cause confusion if and when these version numbers diverge.

**Expected action**
Herbert has made a [proposal ](https://github.com/camaraproject/ReleaseManagement/issues/94) in Release Management to adopt the same release notation as is used for sub-projects, decoupling the release version number from the separate version number (if any) of the included documents and artifacts. If this proposal was adopted, then the next release (likely an "alpha") would be named r2.1, with subsequent versions resulting in a minor version bump until the final version for the next meta-release is declared.

This proposal should be discussed within Commonalities

**Additional context**
API (sub-project) release numbering rules are documented [here](https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/14559630/API+Release+Process)

rartych:
#### What type of PR is this?
* subproject management

Adds VERSION.yaml file for versioning of Commonalities following Release Management gudelines:
https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/14551399/Meta-release+Process#Version-and-release-numbering

Fixes #307

- [ ] Yes
- [x] No

In the file the version is set to `wip`. Should it be maintained the same as API version?

#### Changelog input

```
Added VERSION.yaml file

```

### DIFF
--- a/VERSION.yaml
+++ b/VERSION.yaml
@@ -1,0 +1,1 @@
+version: wip


PedroDiez (approved):
LGTM

patrice-conil (approved):
LGTM

PR #358, opened 2024-12-16
Update rules when using POST for sensitive data

### LINKED ISSUE #247: Request body is required but all properties are optional
**Problem description**
A small one but need global alignement in all our API.
This one was raised by @fernandopradocabrillo in [SIM Swap  API](https://github.com/camaraproject/SimSwap/issues/118) and we got the same question with @FabrizioMoggio in Call Forwarding API.

The request body is required for an api, but all attributes are optional. What should be the body when the client doesn't need to send any attributes?

We have 2 options:

1. Leave it as required and force the client to send an empty body -> `{}`
2. Remove the requirement (`required=true` for the `requestBody`) and let the client send a POST without body

**Expected action**
Define the behaviour when the client doen't need to send any request body param. It could a sentence in the design guideline.
As of now with @fernandopradocabrillo we have a preference for option 1 in sim-swap.

**Additional context**
It is important to define this behavior to include the scenarios in the test plan.

eric-murray:
#### What type of PR is this?
* enhancement/feature

The existing design guidelines allow the request body to be optional when using `POST` to transfer sensitive data. The changes mandate that the request body be present, even if no parameters need to be sent (e.g. if a 3-legged access token is used). In this case, and empty JSON object must be sent.

This change ensures that all APIs passing sensitive data using `POST` adopt a common design pattern.

Fixes #247

- [X] Yes
- [ ] No

APIs with an optional `POST` request body must now make it mandatory, even if no parameters need to be sent

Note: I'm not aware of any current APIs that will be impacted by this change. Device Identifier has a PR which will make the request body mandatory.

None

#### Changelog input

```
 release-note
 - Update rules when using POST for sensitive data making request body mandatory
```

#### Additional documentation 
None

### DIFF
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -301,7 +301,9 @@ In such cases, it is recommended to use one of the following methods to transfer
 - When using `GET`, transfer the data using headers, which are not routinely logged or cached
 - Use `POST` instead of `GET`, with the sensitive data being embedded in the request body which, again, is not routinely logged or cached 
 
-When the `POST` method is used, the resource in the path *must* be a verb (e.g. `retrieve-location` and not `location`) to differentiate from an actual resource creation.
+When the `POST` method is used:
+- the resource in the path MUST be a verb (e.g. `retrieve-location` and not `location`) to differentiate from an actual resource creation
+- the request body itself MUST be a JSON object and MUST be required, even if all properties are optional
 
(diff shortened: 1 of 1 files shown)


PedroDiez (approved):
LGTM

rartych (approved):
LGTM

patrice-conil (approved):
LGTM

PR #373, opened 2025-01-13
String pattern added to x-correlator scheme

### LINKED ISSUE #352: Improve definition of x-correlator header
**Problem description**
For security reasons it is advisable to know the format of any exchanged data. The x-correlator header is just defined as "string". There is no clear definition of how this string is composed, hence it is not possible to perform any check on this header. 
That's a something the security teams are trying to avoid.

**Possible evolution**
My proposal is to define this header in a more detailed way. My first choice is to completely define a format for this header. 
We can decide to use UUID, hence checks can be performed verifying the compliancy of the received string to UUID format

**Alternative solution**
Another solution could be let the liberty to the developer to choose whatever string it wants, maybe jut suggesting to use UUID. But in that case it is needed to at least define a standard maximum length and set of allowed characters.

I am interested in your feedback.

rartych:
#### What type of PR is this?
* correction

String pattern added to x-correlator scheme to improve security and inline with current recommendation for string parameters.
The proposed pattern covers recommended UUID strings and the length of 55 allows to use strings similar to  `traceparent` from Trace Context.

CAMARA_common.yaml was updated accordingly.

Fixes #352 

- [x] Yes
- [ ] No

The change should not be breaking for current implementations using UUID in x-correlator.

#### Changelog input

```
String pattern added to x-correlator scheme

```

### DIFF
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -19,6 +19,7 @@ components:
       description: Correlation id for the different services
       schema:
         type: string
+        pattern: ^[a-zA-Z0-9-]{1,55}$
   parameters:
     x-correlator:
       name: x-correlator
@@ -26,6 +27,7 @@ components:
       description: Correlation id for the different services
       schema:
         type: string
+        pattern: ^[a-zA-Z0-9-]{1,55}$
   schemas:
     TimePeriod:
       properties:
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -28,6 +28,7 @@ components:
       schema:
         type: string
         pattern: ^[a-zA-Z0-9-]{1,55}$
+        example: "b4333c46-49c0-4f62-80d7-f0ef930f1c46"
   schemas:
     TimePeriod:
       properties:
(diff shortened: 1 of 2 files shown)


eric-murray (approved):

PedroDiez (approved):
LGTM

patrice-conil (approved):
LGTM

hdamker:
> Does this PR introduce a breaking change?
> 
> x Yes
>  No

I'm not sure that we should introduce this as a "breaking change" ... if we do so then all 1.0.0 API versions from Fall24 will go to automatically to 2.x.x in Spring25. 

I formulated my [proposal](https://github.com/camaraproject/Commonalities/issues/352#issuecomment-2553827091) along the [vendor opaque value](https://www.w3.org/TR/trace-context/#value) in Trace Context, to avoid that we exclude any valid vendor specific correlator by this change and therefore don't need to declare it as breaking change. But maybe too late here ...

rartych:
Formally it is a breaking change (modifying mandatory parameter).
API Design Guidelines recommend using UUID (if API Client implements "UUID-like" correlator then there is no breaking change , however this recommendation is not included in API definitions - Possibly it can be added in x-correlator description field in CAMARA_common.yaml

Issue #365, opened 2024-12-23
Should error 429 (TOO_MANY_REQUESTS) be mandatory in CAMARA specifications?


rartych:
**Problem description**
The changes introduced in https://github.com/camaraproject/Commonalities/pull/329/files require to define Error status 429 TOO_MANY_REQUESTS in every API specification.

The purpose of it is mainly to inform about exceeded quota limits for given API Consumer.
 
**Expected action**
Verify the requirement - if this error status should be included in every specification.

**Additional context**
Error status 429 can be used also used by API platform to inform about overload conditions.  
Discussed by [TSC](https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/55804112/2024-12-19+TSC+Minutes).

Kevsy:
HI @rartych - I can't see a 429 discussion in that TSC minutes link, please can you check?

> Error status 429 can be used also used by API platform to inform about overload conditions.

This seems to be not aligned with RFC6585 sections [4](https://www.rfc-editor.org/rfc/rfc6585#section-4) and [7.2](https://www.rfc-editor.org/rfc/rfc6585#section-7.2):
-  Section 4 defines 429 as rate limiting per user, not for an overall (all-user) server overload.
- Section  7.2 warns that `When a server is under attack or just receiving a very large number
   of requests from a single party, responding to each with a 429 status
   code will consume resources.    Therefore, servers are not required to use the 429 status code; when
   limiting resource usage, it may be more appropriate to just drop connections, or take other steps.`
   
 Rather, `503 Unavailable` would seem to be the appropriate response to indicate server overload, and keep `429` for per-user rate limiting as defined in the standard.

rartych:
@Kevsy  Thank you for clarification.

I have prepared table comparing how Error 429 was defined in Fall24 meta-release: https://lf-camaraproject.atlassian.net/wiki/x/0gC_Aw

For Spring25 meta-release we need guidelines about removing GENERIC_429_QUOTA_EXCEEDED (as non mandatory) or keeping it according to API functionality.
Subprojects need to reshape error definitions to use enums anyway for the release preparation: https://lf-camaraproject.atlassian.net/wiki/x/AYC3Aw

PedroDiez:
Our view on that is that 429 code can be documented as non-mandatory taking advantage of ongoing PR https://github.com/camaraproject/Commonalities/pull/374.

And let subprojects decide on whether this exception is relevant to be documented in API Spec. 
[...]

Issue #246, opened 2024-07-09
Add rules in design guideline about multi-endpoint API design & implementation requirement


bigludo7:
**Problem description**
In discussion #160 and #212 we have long thread of discussion about implementation of multi-endpoint API.

Several opinions are provided but we have not a clear decision/pattern on how we manage this.

1. Do we decide that an API must be **fully** implemented (all endpoints have to be able to provide 2xx response) in order to be "_certified_"* ?
2. (if 1. answer is yes) If we decide that an API must be fully implemented in order to be "certified"  do we add a rule to enforce multi-yaml definition when distinct endpoint tackles independent data? 
3. (if 1.answer is no) If we allow partial implementation do we mandate use of 501 for non implemented endpoint? As suggested in the design guideline and outcome of discussion #160 

Additionally mention of Runtime Restriction API should be probably in the mix for partial implementation but as we cannot assume it will be used by all API provider & consumer.

(* as we have a GSMA certification but not CAMARA certification we can assume here GSMA certification but this point should be discussed with @MarkCornall and @trehman-gsma)

**Expected action**
Get a decision on these questions once for all and write some rules about it in the design guideline document

**Additional context**
@rartych please do not hesitate to provide comments if I wrongly translate commonalities discussion we had July 8th

hdamker:
As of TSC Minutes https://wiki.camaraproject.org/display/CAM/2024-07-18+TSC+Minutes: 

- Do we decide that an API must be fully implemented (all endpoints have to be able to provide 2xx response) in order to be "certified"* ?
  - Tendency to yes and split to have distinct yaml if independent resource [option 1+2]
  - We need an option for an optional endpoint when they are on the same resource (we cannot split-up):
    - We can add only 501 when, for a multi-endpoint working on a same resource, we have optional endpoint
    - Mandatory endpoints must not have 501 defined as a valid response

gmuratk:
Is there a PR in progress for this issue?

hdamker:
@camaraproject/commonalities_codeowners 
 
I recognized that this issue wasn't considered yet when redefining the error responses and reasons within 0.5.0 (at least as far I can see). I propose to add the following text into the API Design Guidelines:

```
[...]

PR #374, opened 2025-01-13
Clarification on non-mandatory error statuses, 429 made non-madatory

### LINKED ISSUE #355: Should errors 405, 415, 500, 501, 502, 503, 504 be listed in CAMARA_common.yaml
**Problem description**
It was agreed in https://github.com/camaraproject/Commonalities/issues/321 not to include some 40X and 50X errors in API definitions.
Errors 405, 415, 500, 501, 502, 503, 504 are present in CAMARA_common.yaml

**Expected behavior**
Remove definitions from CAMARA_common.yaml

**Alternative solution**
Clarification if errors stay in CAMARA_common.yaml.

**Additional context**

rartych:
#### What type of PR is this?
* correction

Clarification proposed in the issue discussion #355.
429 made non-madatory - #365 
501 mandatory, if and only if an endpoint is optional for implementations

Fixes #355 #365 #246

- [ ] Yes
- [x] No

The discussion in #355 concludes to keep the error definitions in CAMARA_common.yaml.
Adding comments to each error schema (saying if it is mandatory) can be considered.

#### Changelog input

```
Added clarification on non-mandatory error statuses

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -803,6 +803,7 @@ In the following, we elaborate on the existing client errors. In particular, we 
 - For the rest of APIs:
   - Error status 401
   - Error status 403
+  - Error status 501, if and only if an endpoint is optional for implementations
 
 NOTE:
 The documentation of non-mandatory error statuses defined in section 6.1 depends on the specific considerations and design of the given API.
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -805,7 +805,11 @@ In the following, we elaborate on the existing client errors. In particular, we 
   - Error status 403
   - Error status 429 TOO_MANY_REQUESTS (For rate limit control)
 
-NOTE: The remaining Error statuses defined in section 6.1 will be documented based on specific considerations for the given API.
+NOTE:
+The documentation of non-mandatory error statuses defined in section 6.1 depends on the specific considerations and design of the given API.
+ - Error statuses 400, 404, 409, 422: These error statuses should be documented based on the API design and the functionality involved. Subproject evaluate the relevance and necessity of including these statuses in API specifications.
+ - Error statuses 405, 406, 410, 412, 415, and 5xx: These error statuses are not documented by default in the API specification. However, they should be included if there is a relevant use case that justifies their documentation. 
+
 
 ### 6.2 Error Responses - Device Object/Phone Number
 
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -807,7 +807,7 @@ In the following, we elaborate on the existing client errors. In particular, we 
 
 NOTE:
 The documentation of non-mandatory error statuses defined in section 6.1 depends on the specific considerations and design of the given API.
- - Error statuses 400, 404, 409, 422: These error statuses should be documented based on the API design and the functionality involved. Subproject evaluate the relevance and necessity of including these statuses in API specifications.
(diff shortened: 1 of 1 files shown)


documentation/API-design-guidelines.md
```diff
@@ -805,7 +805,11 @@ In the following, we elaborate on the existing client errors. In particular, we
   - Error status 403
   - Error status 429 TOO_MANY_REQUESTS (For rate limit control)
 
-NOTE: The remaining Error statuses defined in section 6.1 will be documented based on specific considerations for the given API.
+NOTE:
+The documentation of non-mandatory error statuses defined in section 6.1 depends on the specific considerations and design of the given API.
+ - Error statuses 400, 404, 409, 422: These error statuses should be documented based on the API design and the functionality involved. Subproject evaluate the relevance and necessity of including these statuses in API specifications.
```
patrice-conil, line 810: Thanks @rartych, 
WDYT about "subproject evaluates" or "subprojects evaluate"

PedroDiez (approved):
LGTM

eric-murray (approved):

patrice-conil (approved):
LGTM

PedroDiez (dismissed):
LGTM

patrice-conil (approved):
LGTM

PedroDiez:
I like to propose a wording within the API design guideline in line with [this comment](https://github.com/camaraproject/Commonalities/issues/246#issuecomment-2609527491)

Not include line 886 (indeed 501 is not a mandatory code in the way like 401, and 403 are) and after line 811 include specific section, in this way:

> Special Consideration for 501 code:
•	This code can be only documented in an API endpoint under decision and agreement within a given API Subproject.
•	Following guidelines to be considered: 
o	The use of optional endpoints is discouraged in order to have aligned implementations and deployments
o	When the API Subproyect concludes the endpoint can be optional and after the evaluation of the approach of API splitting in the case of being and independent resource.

I think it respects the spirit targeted for this code and also refelcts the relevant guidelines derived from TSC.
@rartych, @eric-murray, @hdamker please if you can take a look

rartych:
@PedroDiez I am fine with not mixing 501 with 401 and 403.

I propose to extend the point starting in line 806 with the special considerations.
Then it would look like this:

> - Error statuses 405, 406, 410, 412, 415, and 5xx: These error statuses are not documented by default in the API specification. However, they should be included if there is a relevant use case that justifies their documentation
>   - Special Consideration for error 501 NOT IMPLEMENTED to indicate optional endpoint:
>      - The use of optional endpoints is discouraged in order to have aligned implementations
>      - Only for exceptions where an optional endpoint can not be avoided and defining it in separate, atomic API is not possible -  error status 501 should be documented as a valid response

PedroDiez:
> @PedroDiez I am fine with not mixing 501 with 401 and 403.
> 
> I propose to extend the point starting in line 806 with the special considerations. Then it would look like this:
> 
> > * Error statuses 405, 406, 410, 412, 415, and 5xx: These error statuses are not documented by default in the API specification. However, they should be included if there is a relevant use case that justifies their documentation
> >   
> >   * Special Consideration for error 501 NOT IMPLEMENTED to indicate optional endpoint:
> >     
> >     * The use of optional endpoints is discouraged in order to have aligned implementations
> >     * Only for exceptions where an optional endpoint can not be avoided and defining it in separate, atomic API is not possible -  error status 501 should be documented as a valid response

Perfect Rafal!

rartych:
> I propose to extend the point starting in line 806 with the special considerations. Then it would look like this:

Proposed changes have been applied to the PR

PedroDiez (approved):
LGTM

PR #387, opened 2025-01-21
Allow zero-length strings in x-correlator header

### LINKED ISSUE #383: Allow zero-length strings in x-correlator header to avoid unnecessary breaking of API clients
As discussed yesterday within the TSC I see a challenge with the introduction of the pattern in #373. 

API Clients which have added the the x-correlator header (as it is listed in the API definitions) but intentionally or unintentionally left it empty or set it to an empty string (`""`) might break due to this change:

* before the introduction of the pattern these requests were valid and could be treated as if the header was not present (see e.g. comment in the discussion: https://github.com/camaraproject/Commonalities/discussions/366#discussioncomment-11761319) 
* with the introduction of the pattern these requests will (most?) probably rejected, at least the ones who have set it to an empty string, as the introduced restricting pattern requests a minimum length of 1

My view is that API Provider should be liberal and accept NULL (no value) or empty strings within the x-correlator and treat them as if the header was not present. Within the x-correlator pattern we need to accept also strings with zero-lengths for that.

_Originally posted by @hdamker in https://github.com/camaraproject/Commonalities/discussions/366#discussioncomment-11868789_

rartych:
#### What type of PR is this?
* correction

API Provider should be liberal and accept NULL (no value) or empty strings within the x-correlator and treat them as if the header was not present - the pattern was changed to accept zero length of x-correlator string. 
Such header should be not returned in the response.

Fixes #383 

- [ ] Yes
- [x] No

#### Changelog input

```
correction for empty strings within the x-correlator
```

### DIFF
--- a/documentation/API-design-guidelines.md
+++ b/documentation/API-design-guidelines.md
@@ -1046,9 +1046,9 @@ With the aim of standardizing the request observability and traceability process
 
 | Name           | Description                                   | Schema          | Location         | Required by API Consumer | Required in OAS Definition | 	Example                              | 
 |----------------|-----------------------------------------------|----------------------|------------------|--------------------------|----------------------------|----------------------------------------|
-| `x-correlator` | 	Service correlator to make E2E observability | `type: string`  `pattern: ^[a-zA-Z0-9-]{1,55}$`     | Request / Response | No                       | Yes                        | 	b4333c46-49c0-4f62-80d7-f0ef930f1c46 |
+| `x-correlator` | 	Service correlator to make E2E observability | `type: string`  `pattern: ^[a-zA-Z0-9-]{0,55}$`     | Request / Response | No                       | Yes                        | 	b4333c46-49c0-4f62-80d7-f0ef930f1c46 |
 
-When the API Consumer includes the "x-correlator" header in the request, the API provider must include it in the response with the same value that was used in the request. Otherwise, it is optional to include the "x-correlator" header in the response with any valid value. Recommendation is to use UUID for values.
+When the API Consumer includes non-empty "x-correlator" header in the request, the API Provider must include it in the response with the same value that was used in the request. Otherwise, it is optional to include the "x-correlator" header in the response with any valid value. Recommendation is to use UUID for values.
 
 
 In notification scenarios (i.e., POST request sent towards the listener indicated by `sink` address),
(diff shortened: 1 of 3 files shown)


PedroDiez (dismissed):
LGTM

documentation/API-design-guidelines.md
```diff
@@ -1046,9 +1046,9 @@ With the aim of standardizing the request observability and traceability process
...
-| `x-correlator` | 	Service correlator to make E2E observability | `type: string`  `pattern: ^[a-zA-Z0-9-]{1,55}$`     | Request / Response | No                       | Yes                        | 	b4333c46-49c0-4f62-80d7-f0ef930f1c46 |
+| `x-correlator` | 	Service correlator to make E2E observability | `type: string`  `pattern: ^[a-zA-Z0-9-]{0,55}$`     | Request / Response | No                       | Yes                        | 	b4333c46-49c0-4f62-80d7-f0ef930f1c46 |
 
-When the API Consumer includes the "x-correlator" header in the request, the API provider must include it in the response with the same value that was used in the request. Otherwise, it is optional to include the "x-correlator" header in the response with any valid value. Recommendation is to use UUID for values.
+When the API Consumer includes non-empty "x-correlator" header in the request, the API Provider must include it in the response with the same value that was used in the request. Otherwise, it is optional to include the "x-correlator" header in the response with any valid value. Recommendation is to use UUID for values.
```
Kevsy, line 1051: If a zero-length (null) value is supported, then the expected response should also be agreed and documented. Is it:
1) Response will not include an `x-correlator` header
2) Response will include an `x-correlator` header with null value
3) Response will include an `x-correlator` header with server-generated value

I believe #366 / #373 are favouring (1) , in which case the documentation here should add 'If an x-correlator request header is provided with zero-length (null) value, it will be ignored and not included in the response.'

rartych: I've included a hidden recommendation for option (1) stating :

> When the API Consumer includes **non-empty** "x-correlator" header in the request

But I guess it would be clearer to add the proposed sentence:

>  If an x-correlator request header is provided with zero-length (null) value, it will be ignored and not included in the response.

PedroDiez: Even indicated with empty value we also want to support the option to return a value. Our system returns a server-generated value even not provided by the API client. In case API client provides a value it is keep as it is in the response

rartych:
Added missing pattern in event-subscription-template.yaml

PedroDiez (approved):
LGTM

eric-murray (approved):

Issue #436, opened 2025-03-27
Proposal for x-correlator format change


PedroDiez:
**Problem description**
During Spring 25 commonalities a very restrictive pattern was defined for the values supported in the x-correlator header.
https://github.com/camaraproject/Commonalities/blob/r2.3/documentation/API-design-guidelines.md#9-architecture-headers 
 
We have identified that it can be a limitation for the APPs.

- An APP may already be using correlators with a specific format that it wishes to reuse as a correlator in requests to APIs or in request to the authentication server to obtain the access tokens.

- If the APP is forced to use such a restrictive format, it would either need to change the format of its existing correlators or maintain mapping tables between its internal correlators and the correlators used to send its requests to the APIs.

**Possible evolution**
In order to avoid those limitations the following change is proposed:
Establish as a valid correlator any value supported by an HTTP header as defined in the RFC that standardizes the values allowed in HTTP headers: 
RFC7230: https://datatracker.ietf.org/doc/html/rfc7230#section-3.2
RFC9110: https://datatracker.ietf.org/doc/html/rfc9110#section-6.3 // https://datatracker.ietf.org/doc/html/rfc9110#section-5.5

**Alternative solution**

**Additional context**

Kevsy:
Hi @PedroDiez - RFCs 7230/9110 place no constraint on visible characters nor header length.   I think allowing `x-correlator` to be unconstrained introduces the following problems:

- different servers (Apache, nginx etc.) place their own constraints on header lengths. So a long `x-correlator` that works with one CAMARA API provider may not work with another (their server may return `413 Entity too large`)
- unconstrained header fields means additional security vulnerability checks to prevent buffer overflow or SQL injection attacks, etc.
- if we allow `x-correlator` to be any length/format, but later recognise that doing this causes problems, then we risk a non-backward-compatible change for those API Consumers that use `x-correlator` without constraints.

But I agree that the existing limitation can result in the problems you describe:

>An APP may already be using correlators with a specific format that it wishes to reuse as a correlator in requests to APIs or in request to the authentication server to obtain the access tokens.

>If the APP is forced to use such a restrictive format, it would either need to change the format of its existing correlators or maintain mapping tables between its internal correlators and the correlators used to send its requests to the APIs.

[...]

PR #463, opened 2025-05-16
X-Correlator Format Change

### LINKED ISSUE #422: Management of `x-correlator` header when value input by API consumer does not comply the pattern
**Problem description**
During Spring 25 commonalities a pattern has been defined for the `x-correlator` header.
https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#9-architecture-headers

In order to check the pattern and provide a guidance for implementations when the correlator is input by an API Consumer. We should discussed the expected behavior, bearing in mind it is an optional header for the API consumer.

The question is: How to proceed when input `x-correlator` does not comply the pattern?
**Option 1**: Discard by Implementation and optionally return a valid one.
**Option 2**: Discard by Implementation and always generate&return a valid one.
**Option 3**: Generate an error. In such a case, agree on the error to be returned
- Generic 400: 400 - INVALID_ARGUMENT
- Specific 400: 400 - INVALID_X_CORRELATOR (code as an example)

**Possible evolution**
To be discussed. Initial options considered above 

**Alternative solution**
N/A

**Additional context**
The decision taken may have collateral considerations in tests definitions

PedroDiez:
#### What type of PR is this?

* enhancement

Update x-correlator pattern, as commented by the rationale under Issue #436.
Also manages Issue #422, to have an aligned and documented behaviour from Fall25.

Fixes #422, #436

- [ ] Yes
- [X] No

N/A

#### Changelog input

```
 x-correlator format update

```

#### Additional documentation 

N/A

### DIFF
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -19,7 +19,7 @@ components:
       description: Correlation id for the different services
       schema:
         type: string
-        pattern: ^[a-zA-Z0-9-]{0,55}$
+        pattern: ^[a-zA-Z0-9-_:;.\/<>{}]{0,256}$
   parameters:
     x-correlator:
       name: x-correlator
@@ -27,7 +27,7 @@ components:
       description: Correlation id for the different services
       schema:
         type: string
-        pattern: ^[a-zA-Z0-9-]{0,55}$
+        pattern: ^[a-zA-Z0-9-_:;.\/<>{}]{0,256}$
         example: "b4333c46-49c0-4f62-80d7-f0ef930f1c46"
   schemas:
     TimePeriod:
--- a/artifacts/camara-cloudevents/event-subscription-template.yaml
+++ b/artifacts/camara-cloudevents/event-subscription-template.yaml
@@ -238,14 +238,14 @@ components:
       description: Correlation id for the different services
       schema:
         type: string
-        pattern: ^[a-zA-Z0-9-]{0,55}$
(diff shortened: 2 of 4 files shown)


Kevsy:
LGTM - but do we need to document the behaviour when the client sends an `x-correlator` of zero length?

patrice-conil:
> LGTM - but do we need to document the behaviour when the client sends an `x-correlator` of zero length?

In this case, I think that as we return the value provided by the client, we will send an empty string.

Kevsy:
> In this case, I think that as we return the value provided by the client, we will send an empty string.

Yeah...just seems wasteful. I could understand if sending an empty string meant 'please can the network generate a correlator for me and send it back', but I can't see the rationale for sending an empty string if the network is just going to echo it back.

PedroDiez:
> > In this case, I think that as we return the value provided by the client, we will send an empty string.
> 
> Yeah...just seems wasteful. I could understand if sending an empty string meant 'please can the network generate a correlator for me and send it back', but I can't see the rationale for sending an empty string if the network is just going to echo it back.

This point is already covered Kevin, Patrice:
When the API Consumer includes non-empty "x-correlator" header in the request, the API Provider must include it in the response with the same value that was used in the request. **Otherwise, it is optional to include the "x-correlator" header in the response with any valid value.**

It can be an empty string or a valid value, up to the Telco Operator

cesrgzrz (approved):
LGTM

rartych (approved):
LGTM

Kevsy:
Thanks, LGTM

PR #464, opened 2025-05-16
Enhancements Subscription/Notification Architecture for Fall25

### LINKED ISSUE #437: Enhancements for CAMARA Subscription/Notification Architecture
**Problem description**
While designing the implementation of the baseline architecture for Subscription/Notification we have identified several gaps/limitations in Spring25 solution. 
Based on that, we think we can improve the architecture to provide a better experience to our CAMARA/OGW clients, making easier the integration.

The points we have identified so far are:

- Asynchronous subscription Model (Explicit Subscription Model)

> In POST /subscriptions, settle [subscriptionId](https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml#L540) in 202 Accepted response mandatory. Reason is to provide the subscription's resource identifier for a CAMARA client to be able to ask specifically for it (e.g. to check the status). Otherwise it has to perform a query to GET /subscriptions endpoint, but it may be the case of two subscriptions creation very close in time so this way is not 100% deterministic and it can generate problems to the clients to identify the subscription associated to a given request.

> Also, the definition of an event `subscription-begins` with the same purpose as `subscription-ends` but to inform about the begining of the subscription, it is seen as a valuable feature to inform the API client in asynchronous model when the subscription reaches `ACTIVE` status. We can take as a reference the `subscription-ends` model adapting the wording and defining initially the enum value `SUBSCRIPTION_CREATED`.

> Finally, it has been detected that in [event-subscription-template](https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml#L998), it appears 403 - INVALID_TOKEN_CONTEXT exception. It is clear is a typo and no one will use in Spring25 implementations and we can take advantage of this issue to also fix it.

- Improvement in subscription's lifecycle (Explicit Subscription Model)

Image

> It is proposed to add a new event `subscription-updated`, in the same fashion as `subscription-ends` and `subscription-begins`. We are not thinking in an external request so far (Update subscription feature) but internal workflow scenarios that make a subscription enters in status `INACTIVE`. For instance, when for an `ACTIVE` subscription the user rejects the consent for it, procedure that may happen `out-of-band` (e.g. like a user consents center, where he/she manages their transversal privacy settings). In this scenario, the subscription is not deleted but set to `INACTIVE` and if the user grants again the consent it will be again fully operative (i.e. back to `ACTIVE`). Therefore, we propose the definition of this special event `subscription-updated` taking the reference of `subscription-ends` (with the adjustments required) and defining enum values `SUBSCRIPTION_INACTIVE` (status is set `INACTIVE`) and `SUBSCRIPTION_ACTIVE` (status is set to `ACTIVE`) respectively

- Improvement in `sink` checkings. (Both Implicit and Explicit Subscription Model)
In Issue #416 and related PR #418 it was clarified the use of `https` for `sink` when the protocol used for the notification is HTTP.

> It is also proposed to define an schema in `CAMARA_common.yaml` for the sink format (e.g.).

```         
sink:
          type: string
          format: url
          pattern: ^https:\/\/.+$
          description: The address to which events shall be delivered using the selected protocol.
```

> Add that pattern in [event-subscription-template](https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml#L277)

> Define a specific transversal error in API guidelines https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#61-standardized-use-of-camara-error-responses -> 400 - INVALID_SINK, to be returned when sink is not an https endpoint.

> Adhere such exception in APIs which have implicit subscriptions and document in event-subscription-template.yaml for APIs with explicit susbcriptions

- (NEW - 04/JUN) Align [event-subscription-template.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml) to remove 422 - IDENTIFIER_MISMATCH exception (background [Issue#431](https://github.com/camaraproject/Commonalities/issues/431) // [PR#455](https://github.com/camaraproject/Commonalities/pull/455))

**Possible evolution**
Define this solution in the following Documents:
- CAMARA API Design Guidelines (new format)
- [event-subscription-template.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml)

**Alternative solution**
N/A

**Additional context**
N/A

PedroDiez:
#### What type of PR is this?

* enhancement/feature

This PR provides a resolution for topics commented within Issue #437.
Some points has been discussed within #440.

Explanation about PR:
- Decided to not impact `CAMARA-API-Design-Guide.md` and `CAMARA_common.yaml` documentation.
   - Rationale is that the `sink` concept is managed within `CAMARA-API-Event-Subscription-and-Notification-Guide.md` and 
  `event-subscription-template.yaml`
  - Also in `CAMARA-API-Design-Guide.md` it is defined the format: 400 | {{SPECIFIC_CODE}} | {{SPECIFIC_CODE_MESSAGE}}, so 
  implicitly cover the error 400 - INVALID_SINK as well as other errors already documented in the  `event-subscription- 
  template.yaml` (e.g. 400 - INVALID_TOKEN)
  
  NOTE: So, as the relevant documents are `CAMARA-API-Event-Subscription-and-Notification-Guide.md` and 
  `event-subscription-template.yaml` think is better to not duplicate info as implies more maintenance actions and does not 
   provide any additional value. Anyway up to the WG, just to be transparent with the approach.

- Impacted `CAMARA-API-Event-Subscription-and-Notification-Guide.md`. Added new sections:
  - ### 3.2. `subscription-begins` event
  - ### 3.3 `subscription-updates` event

- Impacted  `event-subscription-template.yaml`
  - Added `sink` pattern and specific `400 - INVALID_SINK` error
  - In POST /subscriptions, settle `subscriptionId` as mandatory in response. As per current model definition it also makes it 
  mandatory in DELETE /subscriptions. That also makes sense, to deal with Asynchronous DELETE scenarios and also when the 
  status `DELETED` is managed by a Telco Operator.
  - Define and add support for `subscription-begins` event
  - Define and add support for `subscription-updates` event
  - Fix typo removing unused exception since Spring25 - `403 - INVALID_TOKEN_CONTEXT`
  - (NEW - 04/JUN) Alignment with [PR#455](https://github.com/camaraproject/Commonalities/pull/455) and remove 422 - IDENTIFIER_MISMATCH exception. Commit [738731f22328ecaf6eaa4042e4d0e01fcfc9ed47](https://github.com/camaraproject/Commonalities/pull/464/commits/738731f22328ecaf6eaa4042e4d0e01fcfc9ed47)
 

Fixes #437 

- [X] Yes
- [ ] No

`subscription-ends` event is renamed to `subscription-ended` event.
This is a breaking change in terms of design due to an event type is renamed. 
APIs impacted (i.e. subscription APIs) are in `initial` versions so as the change can be accommodated without impacting major version.

`422 - IDENTIFIER_MISMTACH` exception is removed

N/A

#### Changelog input

```
 Enhancements Subscription/Notification Architecture for Fall25
```

#### Additional documentation 

N/A


artifacts/camara-cloudevents/event-subscription-template.yaml
```diff
@@ -864,6 +935,12 @@ components:
                 status: 400
                 code: INVALID_TOKEN
                 message: Only bearer token is supported
+            GENERIC_400_INVALID_SINK:
+              description: Invalid sink value
+              value:
+                status: 400
+                code: INVALID_SINK
+                message: Only https sink values are supported
```
rartych, line 943: What about more general message:
```suggestion
                message: sink not valid for the specified protocol
```

Currently we have only HTTP as event delivery protocol, but other protocols are considered by Cloudevents.
This error could be raised not only when `https` is not present, but also when the whole provided address is wrong (not resolvable).

rartych:
For me `subscription-update` sounds more natural  (singular noun) as the event contains one change only.
@eric-murray @Kevsy ?

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
```diff
@@ -335,7 +337,40 @@ For consistency across CAMARA APIs, the uniform CloudEvents model must be used w
...
 Note: For operational and troubleshooting purposes it is relevant to accommodate use of `x-correlator` header attribute. API listener implementations have to be ready to support and receive this data.
 
-### 3.2. `subscription-ends` event
+### 3.2. `subscription-begins` event
+
+Specific event notification type "subscription-begins" is defined to inform listener about subscription initiation.
+
+It is used when subscription creation request (POST /subscriptions) is managed asynchronously. For this specific event, the `data` must feature `initiationReason` attribute. 
+
+The following table lists values for `terminationReason` attribute:
```
bigludo7, line 346: Should be` initiationReason` and not `terminationReason` probably

bigludo7 (commented):
One small miscue to fix.

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
PedroDiez: have also fixed some wording and typos

bigludo7 (approved):
LGTM

rartych:
I am not sure if there is clear requirement if API definitions should add `subscription-begins` and  `subscription-updates` - is it optional or mandatory?

other consideration - changing names:
 `subscription-begins` -->  `subscription-started`
 `subscription-updates` -->  `subscription-updated`
 `subscription-ends` -->  `subscription-ended`

Any view on that?

sachinvodafone:
> Hello @rartych considering your questions:
> 
> I'm fine to have it in all CAMARA Subscription API definition but do we consider them as mandatory for an implementation? I mean if an implem did not trigger subscription-begins do we consider it as non CAMARA compliant? Just my perspective but we should not enforce this - only recommend use of them. Happy to hear from the team.
> 
> For the name I'm fine with both so will follow team preference :)

I believe, starting with optional events provides flexibility for providers to adopt them based on their current capabilities but it may result in inconsistent behaviors—such as one provider supporting an event while another does not. This could require clients to adapt to different implementations during integration.

As part of an initial phased approach, keeping events optional makes sense. However, after collecting feedback and assessing the readiness of telecom providers to support these scenarios consistently, we can move towards making these events mandatory for long term.

patrice-conil (dismissed):
LGTM

PedroDiez:
04/JUN: Taking advantage of this PR, have also covered aligment with [PR#455](https://github.com/camaraproject/Commonalities/pull/455) and remove 422 - IDENTIFIER_MISMATCH exception. Commit [738731f22328ecaf6eaa4042e4d0e01fcfc9ed47](https://github.com/camaraproject/Commonalities/pull/464/commits/738731f22328ecaf6eaa4042e4d0e01fcfc9ed47)

cc @rartych, @bigludo7, @patrice-conil, @eric-murray

patrice-conil (approved):
LGTM

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
```diff
@@ -335,9 +337,44 @@ For consistency across CAMARA APIs, the uniform CloudEvents model must be used w
...
+The following table lists values for `updateReason` attribute:
+
+| enum value | update reason |
+| -----------|-------------------- |
+| SUBSCRIPTION_ACTIVE | API server transitioned susbcription status to `ACTIVE` |
+| SUBSCRIPTION_INACTIVE | API server transitioned susbcription status to `INACTIVE` |
+
+Note1: This enumeration is also defined in `event-subscription-template.yaml` (placed in [Commonalities/artifacts/camara-cloudevents](/artifacts/camara-cloudevents) directory).
+
+Note2: The "subscription-updates" notification is not counted in the `subscriptionMaxEvents`. (for example, if a client request set `subscriptionMaxEvents` to 2, it can  receive 2 notifications, besides the notifications sent for "subscription-updates").
```
rartych, line 373:
```suggestion
Note2: The "subscription-updated" notification is not counted in the `subscriptionMaxEvents`. (for example, if a client request set `subscriptionMaxEvents` to 2, it can  receive 2 notifications, besides the notifications sent for "subscription-updates").
```

jlurien (approved):
LGTM

PR #475, opened 2025-06-04
Add Event versioning section to Guide document

### LINKED ISSUE #474: Addition of event versioning section
**Problem description**
Events need to be version to handle changes in their life cycle.

**Expected action**
Add a section on "Event versioning" to the CAMARA-API-Event-Subscription-and-Notification-Guide.md document

Same information is in the wiki here: https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/14557919/API+versioning#Event-versioning

tanjadegroot:
#### What type of PR is this?

* documentation

Add the section on event versioning to the Guide document.

Fixes #474 

- [ ] Yes
- [x ] No

I have moved some of the text out of and under the table as markdown does not support linebreaks in table cells :-( 

#### Additional documentation 

Same text on wiki (same text): https://lf-camaraproject.atlassian.net/wiki/spaces/CAM/pages/14557919/API+versioning#Event-versioning


documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
```diff
@@ -294,6 +295,76 @@ the two requests should be handled independently & autonomously.
...
+
+ [CloudEvent versioning](https://github.com/cloudevents/spec/blob/main/cloudevents/primer.md#versioning-of-cloudevents) states the following (quoted):
+
+- When a CloudEvent's data changes in a backwardly-incompatible way, the value of the type attribute should generally change. The event producer is encouraged to produce both the old event and the new event for some time (potentially forever) in order to avoid disrupting consumers.
+- When a CloudEvent's data changes in a backwardly-compatible way, the value of the type attribute should generally stay the same.
+
+For CAMARA events, its `type` attribute is the base for event versioning as it contains the event version.
+
+- The format of the `type` attribute value is defined by Commonalities in the `CAMARA-API-Event-Subscription-and-Notification-Guide.md` document as: `org.camaraproject.<api-name>.<event-version>.<event-name>`.
```
rartych, line 309: Self-reference to the document is not needed:
```suggestion
- The format of the `type` attribute value is defined as: `org.camaraproject.<api-name>.<event-version>.<event-name>`. 
```

tanjadegroot:
Good catch Rafal :-)

rartych:
>  Should the note on an array value be kept ?

Yes - it will be changed with PR https://github.com/camaraproject/Commonalities/pull/477 - I hope without conflict.

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
```diff
@@ -317,7 +388,7 @@ For consistency across CAMARA APIs, the uniform CloudEvents model must be used w
...
-| type            | string            | a value describing the type of event related to the originating occurrence. For consistency across API we mandate following pattern: `org.camaraproject.<api-name>.<api-version>.<event-name>` with the `api-version` with letter 'v' and the major version like example org.camaraproject.device-roaming-subscriptions.v1.roaming-status                                                                                                                                                        | mandatory     |
+| type            | string            | a value describing the type of event related to the originating occurrence. For consistency across API we mandate following pattern: `org.camaraproject.<api-name>.<event-version>.<event-name>` with the `event-version` with letter 'v' and a simple version number. An example is org.camaraproject.device-roaming-status-subscriptions.v1.roaming-status                                                                                                                                                        | mandatory     |
```
rartych, line 391: Can we add the reference here too?

```suggestion
| type            | string            | a value describing the type of event related to the originating occurrence. For consistency across API we mandate following pattern: `org.camaraproject.<api-name>.<event-version>.<event-name>` with the `event-version` with letter 'v' and a simple version number - see chapter [2.3. Event versioning](#23-event-versioning). An example is org.camaraproject.device-roaming-status-subscriptions.v1.roaming-status                                                                                                                                                        | mandatory     |
```

PedroDiez:
Also aligned some information in https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml#L8.

Replace:

> Note on event name convention: the event type name MUST follow: ``org.camaraproject.<api-name>.<api-version>.<event-name>``

By

> Note on event name convention: the event type name MUST follow: ``org.camaraproject.<api-name>.<event-version>.<event-name>``

PedroDiez (commented):
Fine to me, commented a minor point

tanjadegroot:
> Also align some information in https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml#L8.
> 
> Replace:
> 
> > Note on event name convention: the event type name MUST follow: `org.camaraproject.<api-name>.<api-version>.<event-name>`
> 
> By
> 
> > Note on event name convention: the event type name MUST follow: `org.camaraproject.<api-name>.<event-version>.<event-name>`
> 
> cc @tanjadegroot, @rartych

Fixed - I checked for all occurrences of "api-version" and they are covered now (where applicable)

rartych (approved):
LGTM

hdamker:
> @rartych @hdamker I know we discussed this, but at second thought I think this is the last (good) opportunity to change the place of the version field to put it behind the event-name: api-name.event-name.event-version
> 
> Advantage: Putting the event-version after the event-name may help in reducing any confusion wrt to the API version.
> 
> I checked, and currently we have around 8 subscription APIs (6 in Spring25) and all are still initial APIs so it should not be a big deal if we change now.
> 
> We can leave the API updates to be done for M4. I can do a quick update of the PR

* I'm also in principle in favor of doing the change, but won't do it here on the fly without a proper discussion as there are pros and cons for both approaches. I found a lot of evidence that putting event-version after the event-name (aka "action"), but propose to discuss that in a separate issue and pull request
* If we do the change we should use it as an exercise how to introduce such a change without forcing APIs to break backward compatibility just for the sake of this change. The way could be to define it as the target schema, but allow APIs to keep the current structure at least until their next major version (or even until the update the event structure).

As codeowner in QualityOnDemand I would like to keep the v1.0.0 API and v1 "event family" stable, and change to a new event versioning approach only later (or not at all as long there are no other changes to the event data structure). We don't have the alternative to offer both event types in parallel unfortunately.

tanjadegroot:
> > @rartych @hdamker I know we discussed this, but at second thought I think this is the last (good) opportunity to change the place of the version field to put it behind the event-name: api-name.event-name.event-version
> > Advantage: Putting the event-version after the event-name may help in reducing any confusion wrt to the API version.
> > I checked, and currently we have around 8 subscription APIs (6 in Spring25) and all are still initial APIs so it should not be a big deal if we change now.
> > We can leave the API updates to be done for M4. I can do a quick update of the PR
> 
> * I'm also in principle in favor of doing the change, but won't do it here on the fly without a proper discussion as there are pros and cons for both approaches. I found a lot of evidence that putting event-version after the event-name (aka "action"), but propose to discuss that in a separate issue and pull request
> * If we do the change we should use it as an exercise how to introduce such a change without forcing APIs to break backward compatibility just for the sake of this change. The way could be to define it as the target schema, but allow APIs to keep the current structure at least until their next major version (or even until the update the event structure).
> 
> As codeowner in QualityOnDemand I would like to keep the v1.0.0 API and v1 "event family" stable, and change to a new event versioning approach only later (or not at all as long there are no other changes to the event data structure). We don't have the alternative to offer both event types in parallel unfortunately.

OK, actually, I think one could introduce the change by adding a new event type (with the same version in a different place) in a backward compatible way in a next release, and and deprecating the old one 2 releases later (as per the standard defined procedure).

OK, fine to take time for more discussion on this.
And please don't sneak in the "F" word again :-) (ref. "family" in your text above. just "v1 events" can do).

/LGTM as is then.

hdamker:
> OK, actually, I think one could introduce the change by adding a new event type (with the same version in a different place) and a copy of the same event structure in a backward compatible way in a next release, and and deprecating the old one 2 releases later (as per the standard defined procedure).

That's unfortunately does not work with implicit subscriptions like in quality-on-demand.

> And please don't sneak in the "F" word again :-) (ref. "family" in your text above. just "v1 events" can do).
Sorry ... I guess we have called it an "event set" in the discussion.

patrice-conil (approved):
LGTM

PR #490, opened 2025-06-27
Clarify that the DeviceResponse schema is not mandatory

### LINKED ISSUE #489: Clarify that the DeviceResponse schema is not mandatory
**Problem description**
The Design Guide states that the `DeviceResponse` schema should be used to disambiguate which device identifier is being used by an implementation when an API consumer has provided more than one identifier, but it does not mandate it. Nor should it, as there will be APIs for which it is not appropriate. If it is used by an API, however, then it becomes mandatory to provide a single device identifier in the response when multiple are provided in the request.

The current documentation does not make this completely clear.

**Expected action**
Clarify the documentation

**Additional context**
Many APIs do not currently allow a device to be included in the response, and forcing them to add one might lead to an overly-complicated response definition

eric-murray:
#### What type of PR is this?
* documentation

The Design Guide states that the `DeviceResponse` schema should be used to disambiguate which device identifier is being used by an implementation when an API consumer has provided more than one identifier, but it does not mandate it. Nor should it, as there will be APIs for which it is not appropriate. If it is used by an API, however, then it becomes mandatory to provide a single device identifier in the response when multiple are provided in the response.

This PR clarifies that:
- The `DeviceResponse` schema is not mandatory for an API
- If used in a response definition, however, the API provider:
  - MUST include in the response which device identifier is being used when multiple have been provided
  - MAY always include the device identifier in the response even when only one has been provided

Fixes #489

- [ ] Yes
- [X] No

None

#### Changelog input

```
 release-note
 - Clarify that the DeviceResponse schema is not mandatory
```

#### Additional documentation 
None


patrice-conil (approved):
LGTM

PedroDiez (approved):
LGTM

PR #470, opened 2025-05-28
Add event-notification feature file template

bigludo7:
#### What type of PR is this?

* documentation

#### What this PR does / why we need it:

This is a proposal of Subscription feature template.
I used the nice work done by @maxl2287 in several subscriptions API ;)

#### Which issue(s) this PR fixes:

Fixes #293 

#### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

#### Changelog input

```
 release-note
- add a template for subscription test case.

```

#### Additional documentation 

This section can be blank.

```
docs

```

PR #492, opened 2025-07-01
Fall 25 Feature Files to provide schema context

### LINKED ISSUE #481: test: all `.feature` files need to provide schema context
**Problem description**

PR #480 provides a reference for the `XCorrelator` definition :

`And the header "x-correlator" complies with the schema at "#/components/schemas/XCorrelator"`

"#/components/schemas/XCorrelator" is a relative URI, so we need some context to resolve it. @bigludo7 added [this explanatory text](https://github.com/bigludo7/Commonalities/blob/56915bd3efd221d7cef28bc25abceefe5deef02f/artifacts/camara-cloudevents/event-subscription-template.feature#L16) following @jlurien's [good comment](https://github.com/camaraproject/Commonalities/pull/470#discussion_r2124338953) in #470:

```
  # References to OAS spec schemas refer to schemas specified in <xxx>-subscriptions.yaml
  # References to schemas starting with the # symbol are JSON Pointers from the root of the OAS document: <xxx>-subscriptions.yaml, Schema names are aligned with the event-subscription-template.yaml artifact.
  

```

...which is fine, but: this is only stated in the event subscriptions template. It is not in API Guidelines, which only states:

```
Background:
    Given an environment at "apiRoot" 
    And the resource "{path_resource}"                                                              |
    And the header "Content-Type" is set to "application/json"
    And the header "Authorization" is set to a valid access token
    And the header "x-correlator" complies with the schema at "#/components/schemas/XCorrelator"
```

 So the problem is that "#/components/schemas/XCorrelator" has no context here: the schema file to which it refers is not declared.

**Expected behavior**

As has been done for the even subscriptions feature template, declare the schema file used in all `.feature` files, either in API Guidelines documentation) or as an explicit declaration in each `.feature` file, e.g.:

`    And the schema definition is located at {API definition.yaml}`

**Alternative solution**

**Additional context**

PedroDiez:
#### What type of PR is this?

* enhancement
* documentation
* tests

This PR manages the topic raised in Issue #481, which indicates the lack of establising an schema context as indicated in PR #470 for event notification transversal gherkin template.

In this way, this PR proposes the following:
- Update [C01-device-errors.feature](https://github.com/camaraproject/Commonalities/blob/main/artifacts/testing/C01-device-errors.feature) and [C02-phoneNumber-errors.feature](https://github.com/camaraproject/Commonalities/blob/main/artifacts/testing/C02-phoneNumber-errors.feature) artifacts to provide such schema context and align with PR #470

- Update [API-Testing-Guidelines.md](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-Testing-Guidelines.md) guide to document the `feature context` already being used across Feature files also referencing to the schema context

Fixes #481 

- [ ] Yes
- [x] No

#### Changelog input

```
 Setting a feature context

```

#### Additional documentation 

N/A

### DIFF
--- a/documentation/API-Testing-Guidelines.md
+++ b/documentation/API-Testing-Guidelines.md
@@ -129,7 +129,9 @@ Feature: CAMARA Device location verification API, v0.2.0 - Operation verifyLocat
 
 #### Feature Context
 
-Every feature will include a context after the `Feature` tag to provide relevant information about the implementation and execution of the tests.
+Every feature may include a context after the `Feature` tag to provide relevant information about the implementation and execution of the tests.
+
+This feature context is RECOMMENDED unless it does not provide additional value (e.g. all the info could have been given in the Background section for a certain API)
 
 For the feature context, the following template should be used:
 
@@ -142,7 +144,25 @@ For the feature context, the following template should be used:
   # Testing assets:
   # * 
   #
-  # References to OAS spec schemas refer to schemas specifies in {apiname}.yaml, version {version}
+  # References to OAS spec schemas refer to schemas specifies in {apiname}.yaml
+```
+
+An example for this feature context is depicted below:
+
+```
+    # Input to be provided by the implementation to the tester
+    #
+    # Implementation indications:
+    # * apiRoot: API root of the server URL
+    # * List of device identifier types which are not supported, among: phoneNumber, ipv4Address, ipv6Address.
+    #   For this version, CAMARA does not allow the use of networkAccessIdentifier, so it is considered by default as not supported.
+    # * List of application server IP formats which are not supported, among ipv4 and ipv6.
+    #
+    # Testing assets:
+    # * A device object applicable for Quality On Demand service.
+    # * A device object identifying a device commercialized by the implementation for which the service is not applicable, if any.
+    #
+    # References to OAS spec schemas refer to schemas specifies in quality-on-demand.yaml
 ```
 
 ### Environment variables
--- a/documentation/API-Testing-Guidelines.md
+++ b/documentation/API-Testing-Guidelines.md
@@ -127,6 +127,24 @@ For the Feature description, API name and version must be included. When the fea
 Feature: CAMARA Device location verification API, v0.2.0 - Operation verifyLocation
 ```
 
+#### Feature Context
+
+Every feature will include a context after the `Feature` tag to provide relevant information about the implementation and execution of the tests.
+
+For the feature context, the following template should be used:
+
+```
+  # Input to be provided by the implementation to the tester
+  #
+  # Implementation indications:
+  # * 
+  #
+  # Testing assets:
+  # * 
+  #
+  # References to OAS spec schemas refer to schemas specifies in {apiname}.yaml, version {version}
+```
+
 ### Environment variables
 
 Commonly, some values to fill the request bodies will not be known in advance and cannot be specified as part of the feature file, as they will be specific for the test environment, and they will have to be provided by the implementation to the tester as a separate set of environment or configuration variables.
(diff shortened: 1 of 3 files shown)


documentation/API-Testing-Guidelines.md
```diff
@@ -127,6 +127,24 @@ For the Feature description, API name and version must be included. When the fea
 Feature: CAMARA Device location verification API, v0.2.0 - Operation verifyLocation
 ```
 
+#### Feature Context
+
+Every feature will include a context after the `Feature` tag to provide relevant information about the implementation and execution of the tests.
+
+For the feature context, the following template should be used:
```
Kevsy, line 134: Can you give some examples of how this template would be filled? Otherwise a `.feature` author may not be sure what to put in each section. 

We should also decide whether the template always needs to be filled out ,  in case that is not necessary for certain .feature files (e.g. all the info could have been given in the Background assertions for certain APIs)

documentation/API-Testing-Guidelines.md
```diff
@@ -127,6 +127,24 @@ For the Feature description, API name and version must be included. When the fea
...
+```
+  # Input to be provided by the implementation to the tester
+  #
+  # Implementation indications:
+  # * 
+  #
+  # Testing assets:
+  # * 
+  #
+  # References to OAS spec schemas refer to schemas specifies in {apiname}.yaml, version {version}
```
jlurien, line 145: As the API version is already indicated at the top of the Feature file, I would not add it again here, as it creates additional maintenance work when creating releases (vX.Y.Z -> vwip -> v....)

documentation/API-Testing-Guidelines.md
PedroDiez: Some examples @Kevsy (I can add it one in the PR). Good point to add an example

Reference: https://github.com/camaraproject/QualityOnDemand/blob/main/code/Test_definitions/quality-on-demand-createSession.feature
```md
    # Input to be provided by the implementation to the tester
    #
    # Implementation indications:
    # * apiRoot: API root of the server URL
    # * List of device identifier types which are not supported, among: phoneNumber, ipv4Address, ipv6Address.
    #   For this version, CAMARA does not allow the use of networkAccessIdentifier, so it is considered by default as not supported.
    # * List of application server IP formats which are not supported, among ipv4 and ipv6.
    #
    # Testing assets:
    # * A device object applicable for Quality On Demand service.
    # * A device object identifying a device commercialized by the implementation for which the service is not applicable, if any.
    #

    # References to OAS spec schemas refer to schemas specifies in quality-on-demand.yaml
```

Reference: https://github.com/camaraproject/DeviceLocation/blob/main/code/Test_definitions/location-retrieval.feature

```md
  # Input to be provided by the implementation to the tester
  #
  # Implementation indications:
  # * List of device identifier types which are not supported, among: phoneNumber, networkAccessIdentifier, ipv4Address, ipv6Address
  #
  # Testing assets:
  # * A device object which location is known by the network when connected. 2 distinct device are required for some scenario.
  # * A device object identifying a device commercialized by the implementation for which the service is not applicable
  # * A device object which location cannot be provided during test by the network.
  #
  # References to OAS spec schemas refer to schemas specifies in location-retrieval.yaml
```

I can also add a sentence to reflect that:
_We should also decide whether the template always needs to be filled out , in case that is not necessary for certain .feature files (e.g. all the info could have been given in the Background assertions for certain APIs)_

documentation/API-Testing-Guidelines.md
PedroDiez: Yes, good point @jlurien, that is already stated within `Feature` tag so better to avoid duplication and additional maintenance work

Kevsy (approved):
LGTM

PR #517, opened 2025-08-26
ErrorInfo Schema properties alignment

### LINKED ISSUE #515: New ErrorInfo schema alters the expected properties order
**Problem description**
CAMARA_common.yaml in r3.3 updated `ErrorInfo` schema with enhanced descriptions, but changing the order of the properties that appeared in previous versions, and is used in all guidelines and specifications, which was `status` -> `code` -> `message`

**Expected behavior**
The same order should be kept. Even if both schemas are equivalent and examples are still compliant with the new schema, tools (spec viewers, code generators, etc) usually respect the the schema order.

**Alternative solution**
If this is not patched before M4 due to time constraints, WGs should ignore the new order and preserve the original one, updating only the enhanced descriptions.

PedroDiez:
#### What type of PR is this?

* correction

This PR aligns ErrorInfo schema in all CAMARA artifacts according to CAMARA API Design guide: https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Design-Guide.md#3-error-responses

Fixes #515 

- [ ] Yes
- [X] No

Taking advantage of this PR, CAMARA artifacts version are reset after CAMARA Public Release Generation

#### Changelog input

```
 ErrorInfo Schema properties alignment

```

#### Additional documentation 

N/A

### DIFF
--- a/VERSION.yaml
+++ b/VERSION.yaml
@@ -1,1 +1,1 @@
-version: 0.6.0
+version: wip
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -5,7 +5,7 @@ info:
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
-  version: 0.7.0
+  version: wip
   x-camara-commonalities: 0.6
   
 paths: {}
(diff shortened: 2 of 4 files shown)


jlurien (approved):
LGTM

patrice-conil (approved):
LGTM

PR #537, opened 2025-09-29
Fix x-camara-commonalities format

### LINKED ISSUE #528: Fix "x-camara-commonalities" value format to be understood as a string
**Problem description**
As per currently defined in [CAMARA_Common.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/CAMARA_common.yaml#L9) along with the decision to not include the patch starting from v0.5, 0.5 is interpreted as a number instead of a version number (string). That is not correct and should be ammended.

That also may have implications as this field can be also used not only to understand the MetaRelease the API version is related to but also to apply some logic (e.g. transversal 422 exceptions)

**Expected behavior**
Set "x-camara-commonalities" value within quotation marks.
I.e: `x-camara-commonalities: "0.6"`

**Alternative solution**
N/A

**Additional context**
N/A

PedroDiez:
#### What type of PR is this?

* correction

This PR fixes  `x-camara-commonalities` format to be interpreted as a string and not an integer as it considers for its versioning major an minor versions.

NOTE: The value is settled to "wip" as being part of the new Spring26 MetaRelease cycle

Fixes #528 

- [ ] Yes
- [x] No

#### Changelog input

```
 Fix x-camara-commonalities format as a string

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/artifacts/CAMARA_common.yaml
+++ b/artifacts/CAMARA_common.yaml
@@ -6,7 +6,7 @@ info:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
   version: wip
-  x-camara-commonalities: 0.6
+  x-camara-commonalities: "wip"
   
 paths: {}
 components:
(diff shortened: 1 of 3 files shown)


patrice-conil (approved):
LGTM

jlurien (dismissed):

PR #550, opened 2025-10-19
New 409 error code called INCOMPATIBLE_STATE

tlohmar:
#### What type of PR is this?

* enhancement/feature

The PR adds a new 409 Error Code called INCOMPATIBLE_STATE, indicating that the referenced resource in the request is in a wrong state for the requested operation. 

Fixes #

- [ ] Yes
- [x] No

#### Changelog input

```

New 409 Error Code called INCOMPATIBLE_STATE, indicating that the referenced resource in the request is in an incompatible state. 

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/documentation/CAMARA-API-Design-Guide.md
+++ b/documentation/CAMARA-API-Design-Guide.md
@@ -278,7 +278,8 @@ In the following, we elaborate on the existing client errors. In particular, we 
 |       409        |        `ABORTED`        | Concurrency conflict.                                               | Concurrency of processes of the same nature/scope                                                                               |
 |       409        |    `ALREADY_EXISTS`     | The resource that a client tried to create already exists.          | Trying to create an existing resource                                                                                           |
 |       409        |       `CONFLICT`        | A specified resource duplicate entry found.                         | Duplication of an existing resource                                                                                             |
-|       409        |       `INCOMPATIBLE_STATE`     | Resource state conflict. | A resource referenced in the request is in an incompatible state for the requested operation                                                                                             |
+|       409        |       `INCOMPATIBLE_STATE`     | A referenced resource is in an incompatible state. | A resource referenced in the request is in an incompatible state for the requested operation                                                                                             |
+
 |       409        |   `{{SPECIFIC_CODE}}`   | `{{SPECIFIC_CODE_MESSAGE}}`                                         | Specific conflict situation that is relevant in the context of the API                                                          |
 
 <font size="3"><span style="color: blue"> Service Exceptions </span></font>


Kevsy:
Well... 

> indicating that the referenced resource in the request is in a wrong state for the requested operation.

...that is the same meaning defined in [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-409-conflict) for 409 Conflict:

> The 409 (Conflict) status code indicates that the request could not be completed due to a conflict with the current state of the target resource.

 Unfortunately, the existing `common.yaml` has redefined 'CONFLICT' to specifically mean 'duplicate entry'

```yaml 
             status: 409
                code: CONFLICT
                message: A specified resource duplicate entry found.
 ```
 
To avoid confusion with the standard, it may be better to align the existing CAMARA definition for CONFLICT with the RFC. Then  if we specifically need an additional 'duplicate entry' error we can see if it should be a `409` or `422` code.

**Review by hdamker — DISMISSED** [review]
Currently the changelog entry in the PR description does not fit to the PR content ("WRONG_STATE" vs "INCOMPATIBLE_STATE"). As we want to move towards more automation, this entry get's important.

Beyond that I would have the question if there is enough evidence that his is a traversal error code which is needed by more than one or two APIs. Otherwise we have in a few months a similar discussion as in #562. I couldn't find an issue for this PR where this was discussed.

**Comment by hdamker** [comment]
@tlohmar 

> Currently the changelog entry in the PR description does not fit to the PR content ("WRONG_STATE" vs "INCOMPATIBLE_STATE"). As we want to move towards more automation, this entry get's important.
> 
Thanks for update the changelog entry. You can consider to edit also the PR title.

> Beyond that I would have the question if there is enough evidence that his is a traversal error code which is needed by more than one or two APIs. Otherwise we have in a few months a similar discussion as in #562. I couldn't find an issue for this PR where this was discussed.

I will dismiss my blocking request for changes, as my analysis (cf. https://github.com/camaraproject/Commonalities/issues/562#issuecomment-3590191272) has shown that this new error code can replace several API specific error codes. But one prerequisite for it is that it is not restricted to referenced resources but also used if there is a conflicting state of the target resource. We can apply this change of your proposal after we have concluded the discussion in #562. 

The proposed code in Commonalities could be something like (description not binding for APIs, need to be adapted otherwise the actual error):

```
          GENERIC_409_INCOMPATIBLE_STATE:
            description: |
              Resource (target or referenced) is in incompatible state for the requested operation.
              Use for both:
              - Target resource state conflicts (e.g., session not in AVAILABLE state for extension)
              - Referenced resource state conflicts (e.g., device already has active conflicting session)
              Message should clarify which resource and required state.
            value:
              status: 409
              code: INCOMPATIBLE_STATE
              message: "Resource is in an incompatible state for this operation"
```

PR #557, opened 2025-11-04
Limited Success Response

### LINKED ISSUE #555: Enhancement Proposal – Standardizing Limited Success Response Handling
**Problem description**
The current **CAMARA API Design Guide** does not explicitly define or standardize handling for *limited success* response cases (informative 2xx responses). These responses represent successful operations with partial fulfillment due to factors such as user preferences or temporary subject unavailability.

As a result, CAMARA APIs currently adopt inconsistent patterns when reporting scenarios like privacy control limitations or subject context restrictions. This inconsistency complicates both client implementation and interoperability across APIs that need to express partial but valid outcomes.

**Possible evolution**
Enhance the **CAMARA API Design Guide** to define and standardize a **Success-Limited Payload** model that can be reused across all APIs. The enhancement would:

- Introduce a unified schema to represent *limited success* outcomes under `CAMARA_common.yaml`.
- Define an explicit discriminator type for limited success cases:
  - **`SUBJECT_CONTEXT`** — indicates a limitation based on contextual or operational factors (e.g., device unreachable, not processable etc.).
- Provide guidance on client handling logic for each type (e.g., retry on SUBJECT_CONTEXT).
- Document example 200 responses for these cases.

This addition would ensure all APIs follow a consistent and auditable structure when returning success-limited responses.

**Alternative solution**
Each API could define its own approach to represent limited success conditions within its schema definitions. However, this would lead to divergence across subprojects and reduced interoperability. Additionally, such ad-hoc implementations would make automated client generation and validation more complex.

By introducing a standardized model and adding a normative section (proposed as **2.3 Limited Success Responses**) in the Design Guide, CAMARA can provide a unified pattern that is simple to extend and maintain.

**Additional context**
This issue proposes an enhancement to the **CAMARA API Design Guide** and the associated `CAMARA_common.yaml` artifact to include standardized limited success response handling.

Example limited success cases:
- **SUBJECT_CONTEXT** → `notReachable`, `notProcessable`

The goal is to support context-based partial outcomes under a single reusable schema, ensuring consistent interpretation by SDKs and API consumers.

cc @gmuratk

gmuratk:
#### What type of PR is this?

* enhancement/feature

The current CAMARA API Design Guide does not explicitly define or standardize handling for limited success response cases (informative 2xx responses). These responses represent successful operations with partial fulfillment due to factors such as user preferences or temporary subject unavailability.

As a result, CAMARA APIs currently adopt inconsistent patterns when reporting scenarios like privacy control limitations or subject context restrictions. This inconsistency complicates both client implementation and interoperability across APIs that need to express partial but valid outcomes.

Fixes #555 

- [ ] Yes
- [x] No

#### Changelog input

```
 release-note
- Introducing a new section '2.3 Limited Success Responses' to API Design Guide 
- Adding common schema definition SuccessLimitedPayload to CAMARA_common.yaml
```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/documentation/CAMARA-API-Design-Guide.md
+++ b/documentation/CAMARA-API-Design-Guide.md
@@ -11,6 +11,7 @@ This document outlines guidelines for API design within the CAMARA project, appl
 - [2. Data](#2-data)
   - [2.1. Common Data Types](#21-common-data-types)
   - [2.2. Data Definitions](#22-data-definitions)
+  - [2.3. Limited Success Responses](#23-limited-success-responses)
 - [3. Error Responses](#3-error-responses)
   - [3.1. Standardized Use of CAMARA Error Responses](#31-standardized-use-of-camara-error-responses)
   - [3.2. Error Responses - Device Object/Phone Number](#32-error-responses---device-objectphone-number)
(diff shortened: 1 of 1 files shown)


Kevsy:
I can think of two categories of 'limited success' in API fulfillment: requests that apply to a single resource, and requests to collections.

For a single resource:
-  an expectation requested in the operation could not be delivered, but a lower level of expectation could. _Examples: location retrieval that can only be sourced at a lower level of accuracy, Quality of Service that is bound to a lower network performance._

For a collection:
- the request could be fulfilled for some of the resources in the collection, and not for others (i.e. 200/201 for some of the resources, 4xx/5xx for others)
- the expectation requested in the operation could be delivered for some resources, but at a lower level of expectation for others 

In each case, the request has been received, understood, processed and responded to. Which makes it a `200`/`201`/`204` per [RFC 7231](https://www.rfc-editor.org/rfc/rfc7231#page-51). But  we want the developer to known that their expectation has not been fully met, so we can provide a description in the `status` and `message` fields as appropriate.

rartych:
Some more thougts from my side. 
I think we need to clearly define what is _"limited success"_ and what is _"failure"_. The simple way of rating/charging responses can be based on response codes, then 200 "Limited success" response providing only context can be misleading.

- in case described by @Kevsy:  expectation requested in the operation could not be delivered, but a lower level of expectation could. _Examples: location retrieval that can only be sourced at a lower level of accuracy, Quality of Service that is bound to a lower network performance._ - such option should be envisaged in regular API response `200/201/204` - (and possibly in request parameter indicating that lower level of service is acceptable)
- in case the request can not be fulfilled due to factors such as user preferences or temporary subject unavailability as it is not related to client (4xx) should we consider [503 Service Unavailable](https://apidog.com/blog/status-code-503-service-unavailable/)? - it is also used when service depends on external APIs or services (user preferences, device not connected) and 503 is used for temporary situations, the request can be repeated - then the normal ErrorInfo can be applied:
  - Should 422 "LOCATION_VERIFICATION.UNABLE_TO_LOCATE"  be changed to 503 "LOCATION_VERIFICATION.UNABLE_TO_LOCATE"?
  - And transversal errors like 503 DEVICE_NOT_CONNECTED, etc. could be defined

Kevsy:
HI @gmuratk 

> where a request is valid but cannot be fully fulfilled

> qos-provisioning: ASSIGNMENT_UNAVAILABLE (200 response, but assignment not possible)

> population-density-data: PopulationDensityOperationNotCompletedExample (200 response, but operation not completed)

These both look like examples of valid requests which cannot be fulfilled (so 'no success' instead of 'limited success'). Consider instead this example of limited success from population-density-data:

>   - `PART_OF_AREA_NOT_SUPPORTED`: Part of the requested area is outside the MNOs coverage area, the cells outside the coverage area will have property `dataType` with value `NO_DATA`."

For your second point:

> Regarding the use of 422: In cases like Location_Verification.Unable_to_locate, the request is syntactically and semantically correct, but the operation cannot be fulfilled. 

I agree: referring to my comment [here](https://github.com/camaraproject/Commonalities/issues/561#issuecomment-3551991129), this is a service exception. I believe a good way to distinguish between client error and service exception is that a request with a client error will always fail if repeated - which may or may not be the case with a service exception such as  `Location_Verification.Unable_to_locate` . As always, the server can help the client by using the response code/message to help the client decide what to do next (e.g. retry the request, wait for a period and retry, etc.)

Kevsy:
Thanks @gmuratk , understood. 

The use of HTTP status codes to indicate API success (or not) is tricky because HTTP completion is not the same as API fulfillment. Especially the scope of the word 'success':  as the RFC says, "The 200 (OK) status code indicates that the request has succeeded", but 'request' in that sentence simply means the HTTP request - not the API request.

> I wanted to show precedence in the APIs that there are cases where limited or no success cases are responded with 200 status, and it will be best if we can define a standard structure to communicate limited or no success cases, without taking away the option for providers to respond with a 'service exception' as it is done today.

Yes, I think that's important.

PR #576, opened 2026-01-16
docs: add device identifier constraints per #513

### LINKED ISSUE #513: docs: ambiguity in DeviceResponse description
**Problem description**

The 'DeviceResponse` object is described as follows:

> An identifier for the end-user equipment able to connect to the network that the response refers to. This parameter is only returned when the API consumer includes the `device` parameter in their request (i.e. they are using a two-legged access token), and is relevant when more than one device identifier is specified, as only one of those device identifiers is allowed in the response.
> 
>  If the API consumer provides more than one device identifier in their request, the API provider must return a single identifier which is the one they are using to fulfil the request, even if the identifiers do not match the same device. API provider does not perform any logic to validate/correlate that the indicated device identifiers match the same device. No error should be returned if the identifiers are otherwise valid to prevent API consumers correlating different identifiers with a given end user.

It is not clear if the API provider may | should include the `$.device` property if only a single identifier was provided in the request.

**Expected action**

Clarify if the API provider may | should include the `$.device` property if only a single identifier was provided in the request. 

**Additional context**

Kevsy:
Added guidelines for the inclusion of the `$.device` property in API responses based on request conditions, specifying scenarios for both API provider discretion and mandatory inclusion for 2-legged access tokens.

#### What type of PR is this?

* documentation

Clarifies the `.feature` constraints that should be used where the `$.device` object has been included in the request body.

Fixes #513 

- [ ] Yes
- [X] No

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/documentation/API-Testing-Guidelines.md
+++ b/documentation/API-Testing-Guidelines.md
@@ -339,7 +339,7 @@ _Mandatory for any API using 2-legged access tokens_
 ```
 And the response property "$.device" contains a single device identifier that was included in the request
 ```
-Note, the inclusion of "$.device" in the request body should be an exisintg constraint for when 2-legged
+Note, the inclusion of "$.device" in the request body SHOULD be an existing constraint for when 2-legged
 access tokens are used.
 
 


documentation/API-Testing-Guidelines.md
```diff
@@ -324,6 +324,25 @@ Typical response validations:
 - the response property "$.code" is "INVALID_ARGUMENT"
 ```
 
+### Inclusion of `$.device` in response
+
+These constrains are conditional on the inclusion of one or more device identifiers in the request body.
```
Kevsy:
```suggestion
These constraints are conditional on the inclusion of one or more device identifiers in the request body.
```

rartych (approved):
LGTM

rartych (approved):
LGTM

PR #578, opened 2026-01-21
Add business-level outcome guidance to API Design Guide

### LINKED ISSUE #555: Enhancement Proposal – Standardizing Limited Success Response Handling

hdamker:
> **Note to reviewers:**  
> This PR has been **significantly updated** based on the discussion in this thread and follow-up feedback.  
> The current proposal is reflected in the latest version of **Section 3.1** and differs from earlier iterations.  
> Please review the PR based on the current text and the explanation in the following comment: https://github.com/camaraproject/Commonalities/pull/578#issuecomment-3809615958.

### What type of PR is this?

* enhancement / feature

---

### What this PR does / why we need it

This PR updates the **CAMARA API Design Guide** by refining Chapter 3 (“Responses”) and adding Section **3.1 – Business-level Outcomes in Successful Responses**.

CAMARA APIs already return negative, partial, or unknown business outcomes under HTTP 2xx, but model these situations in different legitimate ways (explicit enums, nullability/presence rules, or sometimes 4xx usage). The lack of shared guidance leads to inconsistent API designs.

The proposed guidance:

* acknowledges existing modeling approaches as legitimate patterns,
* describes how business-level outcomes can be expressed clearly in successful (2xx) responses,
* introduces two **optional, standardized refinement fields** (`contextCode`, `contextMessage`) to provide additional machine- and human-readable context,
* clarifies that context fields are supplementary and must not replace outcome semantics,
* refines the 2xx vs 4xx discussion by explaining that better outcome modeling in 2xx responses can reduce the need for 4xx on valid requests.

No new schemas are introduced and no changes are mandated for existing APIs. The guidance is primarily intended for new APIs and new MAJOR versions.

Reviewers are encouraged to focus on the **current text of Section 3.1**, as the proposal has evolved during discussion.

---

### Which issue(s) this PR fixes

Fixes #555

---

### Does this PR introduce a breaking change?

* [ ] Yes
* [x] No

This PR adds documentation guidance only. Individual APIs remain responsible for assessing breaking changes in their own versioning context.

---

### Changelog input

```release-note
Refine API Design Guide guidance for modeling business-level outcomes in successful (2xx) responses
```


documentation/CAMARA-API-Design-Guide.md
```diff
@@ -222,7 +221,89 @@ When IpAddr is used in a payload, the property `objectType` MUST be present to i
...
+An example JSON response using HTTP `200`:
+
+```json
+{
+  "outcome": "NOT_APPLICABLE",
+  "outcomeReason": "REGIONAL_PRIVACY_RESTRICTION",
+  "outcomeMessage": "The requested information could not be disclosed for privacy regulation reasons."
+}
+```
+
```
Kevsy: I like this, my only suggestion is that we can inform the API Consumer if the outcome is temporary (that's not the case in this example, but there may be other cases). The addition of an optional 'retryAfter' field with a time metric would indicate that the outcome _may_ change following a subsequent attempt.

documentation/CAMARA-API-Design-Guide.md
```diff
@@ -222,7 +221,89 @@ When IpAddr is used in a payload, the property `objectType` MUST be present to i
...
+The following principles apply to modeling business-level outcomes in successful responses:
+
+* HTTP `2xx` responses indicate that the request was valid and processed; they **MAY** still represent negative, partial, or unknown business outcomes.
+* Business-level outcomes **SHOULD** be modeled explicitly in the response body rather than inferred from missing data or encoded as errors.
+* APIs that can yield multiple business outcomes **SHOULD** expose a mandatory, typed primary outcome field (for example `status`, `verificationResult`, `availability`, or `responseStatus`) defined as an enum or closed set of values.
+* APIs **MAY** add optional refinement fields derived from the primary outcome field name, such as:
+  * a machine-readable `<baseName>Reason` (typed enum), and
+  * a human-readable `<baseName>Message` (free-text string).
+* HTTP `4xx` status codes **MUST** be reserved for true request errors (invalid input, unsupported identifier, authentication/authorization failure, or contract/configuration mismatches).
```
PedroDiez: In this section, is being considered 422 - SERVICE_NOT_AVAILABLE exception as well?

documentation/CAMARA-API-Design-Guide.md
```diff
@@ -222,7 +221,89 @@ When IpAddr is used in a payload, the property `objectType` MUST be present to i
...
+
+CAMARA APIs may need to return negative, partial, or unknown business outcomes even when a request is syntactically valid, authenticated, authorized, and processed successfully (HTTP 2xx). Without explicit guidance, CAMARA APIs have historically modeled these cases inconsistently, for example using boolean-only results, implicit meaning via empty payloads, or HTTP 4xx codes for data unavailability.
+
+#### 3.1.2. Core Principles
+
+The following principles apply to modeling business-level outcomes in successful responses:
+
+* HTTP `2xx` responses indicate that the request was valid and processed; they **MAY** still represent negative, partial, or unknown business outcomes.
+* Business-level outcomes **SHOULD** be modeled explicitly in the response body rather than inferred from missing data or encoded as errors.
+* APIs that can yield multiple business outcomes **SHOULD** expose a mandatory, typed primary outcome field (for example `status`, `verificationResult`, `availability`, or `responseStatus`) defined as an enum or closed set of values.
```
PedroDiez: Think we should specify a common name. That will be benefit for:
- Avoid name divergences when evaluating this outcome within an API initiative
- Ease automation checks or verifications performed in the future

hdamker:
@PedroDiez @Elisabeth-Mueller @gmuratk @shilpa-padgaonkar 

Let me explain why I have chosen to propose domain API specific primary outcome fields:

* Domain APIs already have established primary outcome fields (verificationResult, qosStatus, status, etc.). Forcing a new generic field name (or wrapper) would either duplicate information, or require breaking renames across multiple APIs.
* A common schema is fine for structure, not for naming. Commonalities can recommend a consistent pattern (`<baseName>`, `<baseName>Reason`, `<baseName>Message`) without mandating a single universal property name. That preserves existing contracts and still improves interoperability.
* Interoperability comes from stable semantics, not identical field names. API Consumers don’t benefit if every API uses `outcome` but with wildly different meaning; they benefit if each API clearly defines its outcome enum and reason codes.
* Making a new schema in CAMARA_common.yaml is a bigger coordination burden (versioning, reuse rules, backward compatibility, and adoption across WGs). A guideline pattern is a lower-friction step that can be adopted incrementally.

We can revisit a reusable schema later once the pattern is proven and stable across several APIs. Also can Sub Projects agree in their domains on a common primary outcome field.

documentation/CAMARA-API-Design-Guide.md
hdamker: @PedroDiez see my comment https://github.com/camaraproject/Commonalities/pull/578#issuecomment-3787085782 below

documentation/CAMARA-API-Design-Guide.md
```diff
@@ -222,7 +221,104 @@ When IpAddr is used in a payload, the property `objectType` MUST be present to i
...
+* APIs SHOULD explicitly define and document how business outcomes are represented in successful responses (for example via explicit outcome enums, documented nullability, or documented presence/absence rules), and MUST NOT rely on undocumented inference from missing data.
+* APIs MAY add optional context fields to provide additional information about the outcome:
+  * `contextCode` — a machine-readable code providing additional context
+  * `contextMessage` — a human-readable explanation providing additional context
+* These optional context fields are additive and MUST NOT be the only way to interpret the business outcome of a successful response.
+* Outcome semantics (success, failure, partial, unknown, not applicable) MUST remain visible via the API's domain-specific response semantics and MUST NOT be moved into `contextCode` or `contextMessage`.
+* HTTP `4xx` status codes SHOULD be reserved for true request errors (invalid input, unsupported identifier, authentication/authorization failure, or contract/configuration mismatches).
```
bigludo7: I've still probably issue for this one when the request the valid but the server is not able to serve it.

This is the case in Device location Family. Take Location Verification, if the MNO is not able to locate for any reason a supposed locate-able device we send a 422 unable to locate. For the WG we moved from 200 (with  "verificationResult": "UNKNOWN")  to this 422 to be way more explicit and easy to understand (and charge btw)

Let le know Herbert if i miss something in your proposal.
cc @jlurien

rartych: It is fair point @bigludo7 .
But having business outcome defined i.e. the 200 response covering limited success we can make CAMARA APIs more consistent.
In Location Verification we have:
> If the location information known to the server is older than the specified `maxAge` or not known at all, an error with code `422 LOCATION_VERIFICATION.UNABLE_TO_FULFILL_MAX_AGE` is sent back,

but in device Roaming Status we can have 200 response with: 
>  **LastStatusTime** : The time when the status was last confirmed to be correct. An older status is more likely to now be **incorrect.**

So for the same device (switched-off for some time) we currently can get different kinds of responses (200 and 422).
Proposed approach leads to more common understanding of such cases.

PedroDiez: Hi Herbert, yes. Let me today for providing them

patrice-conil: > @bigludo7 the guidance is not meant to solve the topic 2xx vs 4xxx finally ... there need always to be domain specific discussions. But describing how a limited business outcome may be expressed as a 2xx outcome allows a balanced discussion within the API domains, and probably will avoid cases where a correct and valid API request will be answered by an error (which always requires extra handling for the API client, even if the client cannot change anything about it). Regarding "easier to charge" ... there are pros and cons for provider and consumer -- we might need to think about the consumer (aka customer) first.

Even though the goal isn't to resolve the 2xx vs. 4xx issue, line 245 clearly favors 2xx and will require modifications to APIs that have opted for 4xx in future major versions. If this leads to greater consistency, why not, but care must be taken not to overcomplicate the work of API consumers.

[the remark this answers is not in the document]
PedroDiez: Ok, understood. Yes I will open an issue for that.

PR #569, opened 2025-12-16
Correcting status code 409 related error information.

tlohmar:
#### What type of PR is this?

* correction

#### What this PR does / why we need it:

The PR addresses the recommendations proposed in Issue #562, including the deprecation of the error code `CONFLICT`. 
The PR also improves the description, how these aligned error codes are related to the `ErrorInfo` Object in the response body. 

#### Which issue(s) this PR fixes:

Fixes https://github.com/camaraproject/Commonalities/issues/562

#### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

#### Changelog input

```
The recommendations proposed in Issue #562, including the deprecation of the error code `CONFLICT` are addressed. 
The description, how these aligned error codes are related to the `ErrorInfo` Object in the response body is improved. 

```

#### Additional documentation 

This section can be blank.

```
docs

```

PR #579, opened 2026-01-22
Correction in table in section 3.1 of API Design Guide

### LINKED ISSUE #562: Duplication of error code semantics, i.e. CONFLICT and ALREADY_EXISTS
**Problem description**

CAMARA commonalities includes two error codes within the `409 CONFLICT` HTTP status code, which are semantically identical.

The error code `CONFLICT` has the same meaning as the error code `ALREADY_EXISTS`. Both address the error condition, that an API operation cannot be executed it leads to a duplication of an existing resource.

**Expected behavior**

Remove or disallow the usage of one of the two error codes. Preferably keep `ALREADY_EXISTS`, since clearer for an API developer and more to the point.

**Alternative solution**

**Additional context**

rartych:
#### What type of PR is this?

* correction

After merging https://github.com/camaraproject/Commonalities/pull/569 the table in section [3.1](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Design-Guide.md#31-standardized-use-of-camara-error-responses) was not rendering correctly.

Fixes #562 

- [ ] Yes
- [x] No

#### Changelog input

```
Table in section 3.1 of API Design Guide corrected

```


PedroDiez (approved):
LGTM

Issue #603, opened 2026-03-24
Restructure artifacts directory for $ref consumption and bundling


hdamker:
**Problem description**

The `artifacts/` directory currently contains template files (`CAMARA_common.yaml`, `notification-as-cloud-event.yaml`, `event-subscription-template.yaml`) in a flat layout. These templates duplicate schemas (ErrorInfo, XCorrelator, CloudEvent, Source, DateTime are defined identically in multiple files) and are structured as standalone APIs rather than as `$ref`-consumable components.

With the move toward `$ref`-based consumption of common schemas (#577) and the bundling model proposed in ReleaseManagement#436, API repositories will adopt a canonical layout:

```text
code/
  API_definitions/   # API source files
  common/            # Cached copy of Commonalities common schemas
```

The `$ref` paths in API source files (e.g., `../common/CAMARA_common.yaml#/components/schemas/ErrorInfo`) are relative to this layout. Currently, the Commonalities templates do not use this layout, so their `$ref` paths differ from what API repos actually use. Templates also duplicate schemas instead of referencing a single authoritative source.

**Possible evolution**

Restructure `artifacts/` so that relative `$ref` paths in templates are identical to those in API repositories:

```text
artifacts/
├── common/                                # Authoritative common schema files
│   ├── CAMARA_common.yaml
│   └── CAMARA_event_common.yaml           # Event-specific schemas (CloudEvent, Source, DateTime)
├── api-templates/                         # Copy-and-adapt API templates (maps to API_definitions)
│   ├── sample-service.yaml                # Request-response pattern
│   ├── sample-implicit-events.yaml        # Implicit subscription (callbacks inline)
│   └── sample-service-subscriptions.yaml  # Explicit subscription management
├── event-templates/                       # Event-type specific templates (maps to Event_definitions)
│   └── sample-event-type.yaml             # Can be added later, for #600 work
└── notification-templates/                # Copy-and-adapt notification callback template
    └── sample-notification.yaml
```

A `$ref` like `../common/CAMARA_common.yaml#/components/schemas/ErrorInfo` works identically in templates and in API repositories — templates serve as living proof that the structure works.

This achieves:

- **Identical `$ref` paths**: Templates use the same relative paths as API repositories. No path adjustment needed when copying a template.
- **No schema duplication**: Templates reference `common/` instead of embedding their own copies of ErrorInfo, XCorrelator, CloudEvent, etc.
[...]

PR #604, opened 2026-03-25
Cloudevents decoupled from EventTypeNotification scheme in Event Subscription Template

### LINKED ISSUE #600: Refactor event definition in Event Subscription Template
**Problem description**
The [current template](https://github.com/camaraproject/Commonalities/blob/main/artifacts/camara-cloudevents/event-subscription-template.yaml) has a single monolithic `CloudEvent` schema that owns the `type` field bound to a specific API's event enum, and includes the discriminator mapping to concrete event schemas. This makes it impossible to reuse CloudEvent across APIs without modification, and impossible to add event groups without editing the shared base.

**Possible evolution**
Decoupling CloudEvent from EventTypeNotification scheme can be done like this:

`CloudEvent` —  CloudEvents 1.0 spec  It knows nothing about what values type can take, what data contains, or which concrete schema should be selected. Any CAMARA API that needs to send a notification starts here, and this schema never needs to change regardless of how many APIs or event types are added.
`ApiNotificationEvent `— extends CloudEvent via `allOf` and does exactly two things: constrains type to its own ApiEventType enum, and owns the discriminator mapping from each enum value to a concrete event schema. The api-name segment in the reverse-DNS string (org.camaraproject.api-name.v0.*) is what makes each type globally unique, so two different API groups can independently define identically-named event types without any collision risk. 
`SubscriptionLifecycleEvent` — includes subscription lifecycle events (subscription-started, subscription-updated, subscription-ended) that are common to every CAMARA API that supports subscriptions. 
`ApiEventType` /  `SubscriptionLifecycleEventType` — Each is a plain string enum scoped to exactly one group. Keeping enums separate from the discriminated schema means the set of valid type strings can be referenced independently 
`NotificationEvent` — A `oneOf` union over all group schemas. It owns no discriminator and applies no constraints of its own — its only job is to enumerate which groups are valid at the callback endpoint.  

**Alternative solution**

**Additional context**
Proposed `CloudEvent` scheme can be imported directly as discussed in https://github.com/camaraproject/Commonalities/issues/577
This proposal is also in-line wirh:
https://github.com/camaraproject/Commonalities/issues/556

rartych:
#### What type of PR is this?

* enhancement/feature

Refactors the notification event schema hierarchy in the subscription template to decouple the CloudEvents 1.0 envelope from CAMARA-specific type constraints and discriminator routing.

The original `CloudEvent` schema mixed up three separate issues — the message format, the event types, and the routing rules — which made it hard to reuse the basic schema for different APIs or to add new event groups without changing shared parts.

This PR introduces a layered schema model:

- **`CloudEvent`** (Commonalities-owned) — pure CloudEvents 1.0 envelope. `type` is a`string`. No CAMARA-specific enum values, no discriminator. This schema is now stable and requires no modification regardless of how many APIs or event types are added.
- **`ApiNotificationEvent`** (API project-owned) — extends `CloudEvent` via `allOf`. Constrains `type` to `ApiEventType` and owns the discriminator mapping for API-specific concrete event schemas. Adding a new event type requires only a new enum value in `ApiEventType` and a new discriminator mapping entry here.
- **`ApiEventType`** (API project-owned) — standalone enum of API-specific event type strings. Separated from the discriminated schema so the valid type strings can be referenced independently (e.g. in `SubscriptionRequest.types`).
- **`SubscriptionLifecycleEvent`** + **`SubscriptionLifecycleEventType`** (Commonalities-owned) — structurally parallel to the API-specific group. Covers `subscription-started`, `subscription-updated`, `subscription-ended`.
- **`NotificationEvent`** (API project-owned) — `oneOf` union used as the callback `requestBody` schema. Adding a new event group requires a single line in the `oneOf`.

As a secondary fix, `SubscriptionRequest.types` and `Subscription.types` now reference `ApiEventType` instead of the previous `SubscriptionEventType` / `EventTypeNotification` enums, which incorrectly included lifecycle event types. Lifecycle events are server-initiated and must not appear as subscribable types in a POST body.

Fixes #600

- [ ] Yes
- [x] No

> The callback `requestBody` schema changes from `CloudEvent` to `NotificationEvent`. For implementations that performed strict schema validation against `CloudEvent` at the callback endpoint, this is a schema reference change but not a wire-format change — valid payloads under the old schema remain valid under the new one. The `SubscriptionRequest.types` / `Subscription.types` change removes lifecycle event types from the subscribable enum; if any implementation accepted (and ignored) lifecycle types in POST body, those values would now be rejected as invalid.

- The layering is designed for split-file use as propose in #603 : `CloudEvent`, `SubscriptionLifecycleEvent`, and `SubscriptionLifecycleEventType` are candidates for `CAMARA_event_common.yaml `; `ApiNotificationEvent`, `ApiEventType`, and `NotificationEvent` belong in the `sample-service-subscriptions.yaml`. **Note:** For  `SubscriptionLifecycleEvent`, and `SubscriptionLifecycleEventType` the special events naming pattern should be reviewed to make them reusable in the future.
- OAS 3.0 discriminators require that the `propertyName` field appear in the schema that declares the discriminator, not only in a parent via `allOf`. The `type` property override in `ApiNotificationEvent` and `SubscriptionLifecycleEvent` satisfies this constraint.
- `NotificationEvent` intentionally carries no discriminator. Without a top-level discriminator, tooling resolves the `oneOf` by evaluating each branch in order. This is correct behaviour — the two groups are distinguished by non-overlapping `type` enum values, so evaluation is unambiguous.
- The `EventApiSpecific1` / `EventApiSpecific2` concrete schemas now extend `ApiNotificationEvent` rather than `CloudEvent` directly, preserving the full inheritance chain and can be placed in `sample-event-type.yaml`.

#### Changelog input

```
refactor: decouple CloudEvent envelope from event type constraints in Event Subscription Template

```

### DIFF
--- a/artifacts/camara-cloudevents/event-subscription-template.yaml
+++ b/artifacts/camara-cloudevents/event-subscription-template.yaml
@@ -21,7 +21,7 @@ externalDocs:
   # {apiRepository} MUST be replaced by the CAMARA Subproject Repository name where the API design based on this template is hosted.
 servers:
   - url: "{apiRoot}/api-name/vx.y"
-    # api-name and version should be valued accordingly to the CAMARA API versioning guidelines, e.g. for a version x.y.z, put v0.y for x=0 or just vx for x>0 in the url.
+- url: "{apiRoot}/api-name/vwip"
     variables:
       apiRoot:
         default: http://localhost:9091
@@ -442,7 +442,7 @@ components:
       type: string
       description: |
         Enum of API-specific event type strings for this API project.
-        The reverse-DNS prefix org.camaraproject.<api-name> makes each value
+        The reverse-DNS prefix `org.camaraproject.<api-name>` makes each value
         globally unique, so two different API groups can independently define
         identically-named event types without any collision risk.
       enum:
@@ -610,7 +610,7 @@ components:
           maxLength: 512
           description: |
             Identifies the event type. CAMARA APIs use reverse-DNS notation:
-            org.camaraproject.<api-name>.<event-version>.<event-name>
+            `org.camaraproject.<api-name>.<event-version>.<event-name>`
             The api-name segment makes each type globally unique across API groups.
         specversion:
           type: string
@@ -629,6 +629,8 @@ components:
           $ref: "#/components/schemas/DateTime"
 
     # ─────────────────────────────────────────────────────────────────────────
+    # API-specific notification event group
+    #
     # Extends CloudEvent via allOf and does exactly two things:
     #   1. Constrains `type` to its own ApiEventType enum
     #   2. Owns the discriminator mapping from each enum value to a concrete schema
--- a/artifacts/camara-cloudevents/event-subscription-template.yaml
+++ b/artifacts/camara-cloudevents/event-subscription-template.yaml
@@ -13,7 +13,7 @@ info:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
   version: wip
-  x-camara-commonalities: "0.7"
+  x-camara-commonalities: 0.7.0
 
 externalDocs:
   description: Product documentation at CAMARA
--- a/artifacts/camara-cloudevents/event-subscription-template.yaml
+++ b/artifacts/camara-cloudevents/event-subscription-template.yaml
@@ -221,10 +221,14 @@ components:
     openId:
       type: openIdConnect
       openIdConnectUrl: https://example.com/.well-known/openid-configuration
+      description: OpenID Connect authentication via discovery metadata.
     notificationsBearerAuth:
       type: http
       scheme: bearer
       bearerFormat: "{$request.body#/sinkCredential.credentialType}"
+      description: |
+        Bearer token for notification delivery. Token format is determined
+        by `sinkCredential.credentialType` in the subscription request.
   parameters:
     SubscriptionId:
       name: subscriptionId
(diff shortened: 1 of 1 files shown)


artifacts/camara-cloudevents/event-subscription-template.yaml
```diff
@@ -424,35 +431,65 @@ components:
...
+    ApiEventType:
       type: string
       description: |
-        event-type - Event triggered when an event-type event occurred. Several event-type could be defined.
-
-        - subscription-started: Event triggered when the subscription starts
-        - subscription-updated: Event triggered when the subscription is updated
-        - subscription-ended: Event triggered when the subscription ends
+        Enum of API-specific event type strings for this API project.
+        The reverse-DNS prefix org.camaraproject.<api-name> makes each value
```
PedroDiez:
```suggestion
        The reverse-DNS prefix `org.camaraproject.<api-name>` makes each value
```

PedroDiez (commented):
Great work @rartych! Descriptive information provided will help a lot in understanding this enhanced model

Just minor editorial suggestions

LGTM at functional level

PedroDiez (approved):
LGTM

patrice-conil (approved):
LGTM

bigludo7 (approved):
Thanks Rafal

hdamker:
> As explained in [#604 (comment)](https://github.com/camaraproject/Commonalities/pull/604#discussion_r3004980258) Github diff is not readable.
> You can see the changes : [here](https://editor.mergely.com/eCx4VSkF)

nice view in "mergely" ... but also readable in GitHub if you hide the whitespace changes: https://github.com/camaraproject/Commonalities/pull/604/changes?w=1 (also available via the UI, at the settings buttons in file changes view).

PR looks good to me as well.

PR #617, opened 2026-04-14
Add pagination shared components and example to artifacts

### LINKED ISSUE #592: Pagination for `GET /resources` operations
**Problem description**
[OWASP linting rules](https://github.com/camaraproject/Commonalities/blob/main/documentation/Linting-rules.md#5-owasp-api-security-top-10-2023-rules) include:
`owasp:api4:2023-array-limit`  that requires schema of type array must specify `maxItems`.

When the expected size of response for `GET /resources` operations is huge it is recommended to use pagination (for API security and efficiency).

The example is [Event subscription template](https://github.com/camaraproject/Commonalities/pull/591)

As currently only one CAMARA API uses pagination, the [current guidelines for pagination](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Design-Guide.md#41-pagination) may need to be reviewed before more often implementation.
Common artifacts can be created to help implementing pagination in API specifications in consistent way.

**Possible evolution**

- Review [current guidelines for pagination](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Design-Guide.md#41-pagination)
- Create artifacts in CAMARA_common.yaml
- apply pagination in Event subscription template for `GET  /subscriptions` operation

**Alternative solution**

**Additional context**

rartych:
#### What type of PR is this?
* enhancement/feature

Adds pagination support to the `GET /subscriptions` in the sample-service-subscriptions API template, following the Commonalities pagination guidelines. 
- Added `page` and `perPage` query parameters to `GET /subscriptions`, referencing new shared `Page` and `PerPage` parameter components
- Changed the `200` response body from an inline `type: array` to a `SubscriptionList` envelope schema (`items[]` + `pagination` object), consistent with the `ResourceList` pattern
- Added `X-Total-Count`, `X-Total-Pages`, and `Link` (RFC 8288) response headers to the `200` response

Clarification on naming and `pagination` object presence added to Design Guide - based on comments from the review

Adds shared components to CAMARA_common.yaml
  - `components/parameters`: `page`, `perPage`
  - `components/headers`: `x-total-count`, `x-total-pages`, `link`
  - `components/schemas`: `Pagination`, `Page`, `PerPage`, `TotalCount`, `TotalPages`

Fixes #592

- [x] Yes
- [ ] No

 The `200` response schema changes from a bare array to an object envelope. Clients consuming the raw array will need to update to read `subscriptions[]`

- `totalCount` and `totalPages` are described as optional in `Pagination` to accommodate endpoints where a full COUNT query is expensive — currently there is no explicit `required:` for  `Page` and `PerPage`.

#### Changelog input

```
Added pagination support to the `GET /subscriptions` in the sample-service-subscriptions API template and shared components to CAMARA_common.yaml.

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/artifacts/api-templates/sample-service-subscriptions.yaml
+++ b/artifacts/api-templates/sample-service-subscriptions.yaml
@@ -1,612 +1,931 @@
-# NOTE: This is a template. Replace all occurrences of "sample-service-subscriptions"
-# (and its Title Case variant) with the actual API name in the appropriate casing.
-openapi: 3.0.3
 info:
-  title: Sample Service Subscription Template
+  title: CAMARA common data types
   description: |
-    This file is a template for CAMARA API explicit subscription endpoint and for Notification model.
-    Additional information are provided in [CAMARA API Event Subscription and Notification Guide](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md).
+    Common data types for CAMARA APIs.
+    This file contains Commonalities-owned schemas that are identical across all
+    CAMARA APIs, including error responses, common parameters, headers, and
+    reusable data types.
 
-    Note on event name convention: the event type name MUST follow: ``org.camaraproject.<api-name>.<event-version>.<event-name>``
-
-    Note on ``security`` - ``openId`` scope: The scope values must be replaced accordingly to the format defined in the guideline document.
-
+    API repositories place this file in `code/common/` and reference schemas
+    via `$ref: "../common/CAMARA_common.yaml#/components/schemas/<SchemaName>"`.
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0.html
   version: wip
   x-camara-commonalities: 0.7.0
 
-externalDocs:
-  description: Product documentation at CAMARA
-  url: https://github.com/camaraproject/apiRepository
-  # {apiRepository} MUST be replaced by the CAMARA Subproject Repository name where the API design based on this template is hosted.
-servers:
-  - url: "{apiRoot}/sample-service-subscriptions/vwip"
-    variables:
-      apiRoot:
-        default: http://localhost:9091
-        description: API root, defined by the service provider, e.g. `api.example.com` or `api.example.com/somepath`
-tags:
-  - name: Sample Service Subscription
-    description: Operations to manage event subscriptions on event-type event
-
-paths:
-  /subscriptions:
-    post:
-      tags:
-        - Sample Service Subscription
-      summary: "Create a sample service event subscription"
-      description: Create a sample service event subscription
-      operationId: createSampleServiceSubscription
-      parameters:
-        - $ref: "../common/CAMARA_common.yaml#/components/parameters/x-correlator"
-      security:
-        - openId:
-            - sample-service-subscriptions:event-type1:grant-level
-            - sample-service-subscriptions:event-type2:grant-level
-      requestBody:
-        content:
-          application/json:
-            schema:
-              $ref: "#/components/schemas/SubscriptionRequest"
-        required: true
-      callbacks:
-        notifications:
-          "{$request.body#/sink}":
-            post:
-              summary: "notifications callback"
-              description: |
-                Important: this endpoint is to be implemented by the API consumer.
-                The sample service server will call this endpoint whenever a sample service event occurs.
-                `operationId` value will have to be replaced accordingly with WG API specific semantic
-              operationId: postNotification
-              parameters:
-                - $ref: "../common/CAMARA_common.yaml#/components/parameters/x-correlator"
-              requestBody:
-                required: true
-                content:
-                  application/cloudevents+json:
-                    schema:
-                      $ref: "#/components/schemas/NotificationEvent"
-              responses:
-                "204":
-                  description: Successful notification
-                  headers:
-                    x-correlator:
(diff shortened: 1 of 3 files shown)


PedroDiez (dismissed):
LGTM

patrice-conil (approved):
LGTM

artifacts/common/CAMARA_common.yaml
```diff
@@ -20,18 +20,70 @@ components:
...
+
   parameters:
     x-correlator:
       name: x-correlator
       in: header
       description: Correlation id for the different services
       schema:
         $ref: "#/components/schemas/XCorrelator"
+
+    page:
```
hdamker, line 67: The Design Guide specifies page default: 1 and perPage default: 20. The shared schemas here have no default keyword. Since these are the reusable components, adding default: 1 and default: 20 would enforce the guideline more precisely.

artifacts/api-templates/sample-service-subscriptions.yaml
```diff
@@ -360,6 +365,22 @@ components:
         id:
           $ref: "../common/CAMARA_event_common.yaml#/components/schemas/SubscriptionId"
 
+    SubscriptionList:
+      description: Response with subscriptions list
+      type: object
+      required:
+        - subscriptions
```
hdamker, line 372: The Design Guide says "the pagination object SHOULD always be present." Making it required in the schema (or at least documenting why it's optional) would be more aligned.

artifacts/common/CAMARA_common.yaml
```diff
@@ -256,6 +308,53 @@ components:
...
+      example: 87
+    TotalPages:
+      type: integer
+      format: int32
+      minimum: 0
+      maximum: 2147483647
+      description: Total number of pages. Equals ceil(totalCount / perPage). MAY be omitted where totalCount is omitted.
+      example: 5
+
+
```
hdamker:
```suggestion

```

artifacts/api-templates/sample-service-subscriptions.yaml
```diff
@@ -360,6 +365,22 @@ components:
...
+      properties:
+        subscriptions:
```
hdamker, line 375: Section 4.1 of the Design Guide uses a generic items property name in the ResourceList pattern. Here subscriptions is used. This is arguably fine for a concrete API (more descriptive), but it deviates from the guideline pattern and may cause inconsistency across APIs. Worth clarifying whether the guideline intends items as the canonical property name or just an example.

artifacts/common/CAMARA_common.yaml
```diff
@@ -256,6 +308,53 @@ components:
...
+      format: int32
+      minimum: 1
+      maximum: 100
+      description: Number of items per page.
+      example: 20
+    TotalCount:
+      type: integer
+      format: int32
+      minimum: 0
+      maximum: 2147483647
```
hdamker, line 348: The Design Guide's example schemas don't specify this. Not wrong, but I guess it is just for the sake of fulfilling the Spectral OWASP rule. That's one of the rules I would questioning ... format int32 already defines the same limit, right?

artifacts/common/CAMARA_common.yaml
```diff
@@ -256,6 +308,53 @@ components:
...
+      description: Total number of items matching the query, after filters applied. MAY be omitted where a full count query is prohibitively expensive.
+      example: 87
+    TotalPages:
+      type: integer
+      format: int32
+      minimum: 0
+      maximum: 2147483647
```
hdamker, line 355: Dito.

hdamker (commented):
The PR looks good to me and is necessary. Some comments for consideration, none of them blocking.

artifacts/api-templates/sample-service-subscriptions.yaml
rartych: It was intended to give just generic example in Design Guide.
The naming should be decided in concrete API definition.

hdamker: I guessed so, maybe it made be clearer the design guide.

artifacts/api-templates/sample-service-subscriptions.yaml
rartych: the pagination object SHOULD always be present when pagination is used.
When response size fits into on page then pagination data is redundant.

artifacts/common/CAMARA_common.yaml
patrice-conil: As Herbert said, int32 sets the maximum, here we restrict to unsigned int32 so we have to keep minimum but we can remove maximum.

artifacts/api-templates/sample-service-subscriptions.yaml
patrice-conil: > the pagination object SHOULD always be present when pagination is used. When response size fits into on page then pagination data is redundant.

As a developer, if I call a paginated API, I expect to always receive the same format whether there is one or multiple pages.

artifacts/common/CAMARA_common.yaml
rartych: The `maximum` attribute is imposed by Spectral OWASP rule

rartych:
Applied changes from review.

patrice-conil (approved):
LGTM

hdamker (approved):

bigludo7 (approved):

PedroDiez (approved):

PR #618, opened 2026-04-15
Empty Request Body Test proposal. Template Artifact for 400 Syntax

### LINKED ISSUE #609: Conflict Behaviour in 2-legged scenario: 400 Empty Request Body vs 422 Missing Identifier
**Problem description**
When reviewing several GSMA collections for API versions in both Spring25 and Fall25 MetaReleases, we have realised that there can be a conflict in 2-legged scenario between tests 400 - Empty Request Body and 422 - Missing Identifier, depending on the logic order designed and implemented by the API Provider.

This situation has been shared with GSMA and they have commented they have received similar comments from other inputs, so this issue aims to propose a means that overcomes this topic. Also it enables a source of truth for GSMA collections used for certifications, supporting GSMA work on it (cc @trehman-gsma, @minpatelgsma).

See below Quality On Demand v1.1 CAMARA tests as an exemplification of involved context:

```
  [@quality_on_demand_createSession_400.3_empty_request_body](https://github.com/camaraproject/QualityOnDemand/blob/r3.2/code/Test_definitions/quality-on-demand-createSession.feature#L201)
  Scenario: Empty object as request body
    Given the request body is set to {}
    When the request "createSession" is sent
    Then the response status code is 400
    And the response header "x-correlator" has same value as the request header "x-correlator"
    And the response header "Content-Type" is "application/json"
    And the response property "$.status" is 400
    And the response property "$.code" is "INVALID_ARGUMENT"

  [@quality_on_demand_createSession_C01.05_missing_device](https://github.com/camaraproject/QualityOnDemand/blob/r3.2/code/Test_definitions/quality-on-demand-createSession.feature#L145)
  Scenario: Device not included and cannot be deduced from the access token
    Given the header "Authorization" is set to a valid access token which does not identify a single device
    And the request body property "$.device" is not included
    When the HTTP "POST" request is sent
    Then the response status code is 422
    And the response property "$.status" is 422
    And the response property "$.code" is "MISSING_IDENTIFIER"
    And the response property "$.message" contains a user-friendly text
```

In a 2 legged token scenario, depending on API Provider logic order implementation designed, the checking of the presence of "device/phoneNumber" can be performed first so as an empty request body "`{}`" implicitly triggers the scenario of 422 - Missing Identifier.

This situation happens in APIs that can work in 2-legged/3-legged scenarios, own the `phoneNumber`/`device` concept and also have other required properties in the request body (1).

From TEF side, in the 2-legged scenario, our business logic prioritizes 422 - Missing Identifier in the case of empty request body "`{}`". 
Rationale under it is the following:

- 422 - UNNECESSARY_IDENTIFIER/MISSING_IDENTIFIER logic is closer to AuthN/AuthZ checkings which is usually performed before syntax logic (i.e. API Provider needs to check token nature (2 legged or 3 legged) in order to identify whether `device`/`phoneNumber` is needed or not)

- Not doing in that way, i.e. applying all syntactic validation logic to the API before 422 UNNECESSARY_IDENTIFIER/MISSING_IDENTIFIER checkings can cause inconsistencies, such as throwing an error in the 3-legged scenario because the phoneNumber doesn't match the pattern or the device object is syntactically invalid ("device": {}), when the relevant issue is that the field shouldn't be sent. In other words, you'd get an error saying "this field is incorrect," and when it's sent correctly, you'd get an error saying "this field shouldn't be sent."

Therefore, currently defined 400 - Empty Request Body test for CAMARA APIs in situation indicated in (1) is wide enough and really covers in 2-legged scenario two situations that a developer MUST consider and deal with:
- Include `phoneNumber`/`device` accordingly to API specification - Fix 422 - MISSING_IDENTIFIER
- Include required request body fields - Fix 400 INVALID_ARGUMENT

**Possible evolution**

We are in a situation where is fulfilled the condition: "if I have two issues in a request I have to solve both" (solving one does not fix the other). Therefore, in this situation CAMARA should define tests that only checks one condition, because with the current test spelling both behaviours can be valid.

Therefore, the proposal would be:
- Not define 400 Empty Request Body in 2-legged (reduce its scope to 3-legged only), due to it cover two casuistics: not fullfilling API specification "contract" as invalid sintax and absence of a needed field (`device`/`phoneNumber`).

> NOTE: We have already design these kind of situations in the opposite way, e.g. [Empty device Test](https://github.com/camaraproject/Commonalities/blob/main/artifacts/testing/C01-device-errors.feature#L32) is not defined for 3-legged case because in this scenario it is not only incurs in a syntax issue but also in the issue that the field is not needed.

Additionally to complement this, a 2-legged "missing mandatory parameter" test could be added (indeed we can also include for the 3-legged case as well), which would cover exactly that case but without simultaneously introducing the error of not sending the `phoneNumber`/`device`.

NOTE: Based on the output/conclusion of this issue, We would also like to take advantage and propose to generate a 400 - Syntax testing artifact to compile this and also document common 400 Errors used among many CAMARA APIs, in the same fashion we have done for phoneNumber/device specific testing templates.

**Alternative solution**
Open discussion in Community to decide a way forward

**Additional context**
N/A

PedroDiez:
#### What type of PR is this?

* enhancement/feature
* tests

This proposal manages to avoid Conflict Behaviour in 2-legged scenario: 400 Empty Request Body vs 422 Missing Identifier, considering an Empty Request Body test only in the case of 3-legged scenario.

Taking advantage of it, a template Artifact for 400 Syntax errors is proposed, in the same fashion as others already documented (device/phoneNumber error scenarios, subscription APIs)

Fixes #609 

- [ ] Yes
- [X] No

#### Changelog input

```
 New common artifact for test plans for 400 Syntax errors. 
 
```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/artifacts/testing/syntax-errors-template.feature
+++ b/artifacts/testing/syntax-errors-template.feature
@@ -25,7 +25,7 @@ Feature: CAMARA Template Artifact 400 - Test scenarios for 400 syntax errors
 
   @{feature_identifier}_{operationId}_400.1_schema_not_compliant
   Scenario: Invalid Argument. Generic Syntax Exception
-    Given the request body is set to any value which is not compliant with the schema at "/components/schemas/<requestSchema>"
+    Given the request body is included but is not compliant with the schema at "/components/schemas/<requestSchema>"
     When the request "{operationId}" is sent
     Then the response status code is 400
     And the response header "x-correlator" has same value as the request header "x-correlator"
--- a/artifacts/testing/syntax-errors-template.feature
+++ b/artifacts/testing/syntax-errors-template.feature
@@ -1,0 +1,118 @@
+Feature: CAMARA Template Artifact 400 - Test scenarios for 400 syntax errors
+
+    CAMARA Commonalities: 0.7.0
+    
+    Common syntax error scenarios for operations.
+
+    NOTES:
+    * This is not a complete feature but a collection of scenarios that can be applied with minor
+    modifications to test plans. Test plans would have to copy and adapt the scenarios as part of
+    their own feature files, along with other scenarios
+
+    * These scenarios assume that other properties not explicitly mentioned in the scenario
+    are set by default to a valid value. This can be specified in the feature Background.
+
+    * {feature_identifier} has to be substituted to the value corresponding to the feature file where
+    these scenarios are included.
+
+    * {operationId} has to be substituted to the value of operationId for the tested operation
+
+  # This feature file is to be used by CAMARA subproject for Common 400 syntax error scenarios for operations.
+  #
+  # References to OAS spec schemas refer to schemas specified in {apiname}.yaml
+
+  # Syntax Error scenarios
+
+  @{feature_identifier}_{operationId}_400.1_schema_not_compliant
+  Scenario: Invalid Argument. Generic Syntax Exception
+    Given the request body is set to any value which is not compliant with the schema at "/components/schemas/<requestSchema>"
+    When the request "{operationId}" is sent
+    Then the response status code is 400
+    And the response header "x-correlator" has same value as the request header "x-correlator"
+    And the response header "Content-Type" is "application/json"
+    And the response property "$.status" is 400
+    And the response property "$.code" is "INVALID_ARGUMENT"
+    And the response property "$.message" contains a user friendly text
+
+  @{feature_identifier}_{operationId}_400.2_no_request_body
+  Scenario: Missing request body
+    Given the request body is not included
+    When the request "{operationId}" is sent
+    Then the response status code is 400
+    And the response header "x-correlator" has same value as the request header "x-correlator"
+    And the response header "Content-Type" is "application/json"
+    And the response property "$.status" is 400
+    And the response property "$.code" is "INVALID_ARGUMENT"
+    And the response property "$.message" contains a user friendly text
+
+  @{feature_identifier}_{operationId}_400.3_empty_request_body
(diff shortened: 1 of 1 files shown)


artifacts/testing/syntax-errors-template.feature
```diff
@@ -0,0 +1,118 @@
...
+
+  # This feature file is to be used by CAMARA subproject for Common 400 syntax error scenarios for operations.
+  #
+  # References to OAS spec schemas refer to schemas specified in {apiname}.yaml
+
+  # Syntax Error scenarios
+
+  @{feature_identifier}_{operationId}_400.1_schema_not_compliant
+  Scenario: Invalid Argument. Generic Syntax Exception
+    Given the request body is set to any value which is not compliant with the schema at "/components/schemas/<requestSchema>"
```
Kevsy: To avoid any overlap with the case where the request body is not included, or is the empty object, maybe best to say:

```suggestion
    Given the request body is included but is not compliant with the schema at "/components/schemas/<requestSchema>"
```

artifacts/testing/syntax-errors-template.feature
```diff
@@ -0,0 +1,118 @@
+Feature: CAMARA Template Artifact 400 - Test scenarios for 400 syntax errors
+
+    CAMARA Commonalities: 0.7.0
+
```
rartych: line with "trailing spaces":
```suggestion

```

rartych (approved):
LGTM

Kevsy (approved):
LGTM

Issue #615, opened 2026-04-13
HTTPSettings.headers in CAMARA_event_common.yaml: remove or constrain?


hdamker:
**Problem description**

`HTTPSettings.headers` in `artifacts/common/CAMARA_event_common.yaml` (lines 170–178, post-#606) is a map of arbitrary custom HTTP headers attached to outbound notification requests. Four observations:

1. **Self-flagged as undiscussed.** The field's own description states: *"Use/Applicability of this concept has not been discussed in Commonalities. When required by an API project as an option to meet a UC/Requirement, please generate an issue for Commonalities discussion about it."*

2. **Not endorsed by the Design Guide.** §5.8.5 standardizes only `x-correlator` (explicitly for notification scenarios) and `Content-Type`. There is no concept of arbitrary custom headers.

3. **No CAMARA API actually uses it.** Across the API spec files in `camaraproject/*` that define an `HTTPSettings` schema, most carry the field verbatim as a placeholder (including the self-flagging note). The one exception surfaced in the checks so far is **SessionInsights**, which defines `HTTPSettings` as `type: object, description: "No additional settings"` — explicitly omitting `headers`. No API populates the field with examples, constraints, or documented use cases.

4. **Triggers two OWASP rules** on every API that `$ref`s this schema:
   - `owasp:api3:2023-constrained-additionalProperties` (no `maxProperties` cap)
   - `owasp:api3:2023-no-additionalProperties` (the field's purpose is `additionalProperties`, so the strict form would break functionality)

**Expected behavior**

Two options for discussion:

**Option A — Remove the `headers` property entirely.** Cleanest fix; honors the field's own "please raise an issue before using" path by acting on the fact that no one has raised one. Leaves `HTTPSettings` with just `method`:

```yaml
    HTTPSettings:
      type: object
      description: HTTP protocol settings for event delivery.
      properties:
        method:
          type: string
          description: The HTTP method to use for sending the message.
          enum:
            - POST
```

Tradeoff: CAMARA drifts further from the CloudEvents subscriptions 0.1-wip spec, which defines `headers` as part of HTTP protocol settings.

[...]

PR #634, opened 2026-05-14
fix(design-guide): add rule for undeclared properties in request bodies

### LINKED ISSUE #632: Design Guide: explicit guidance on additionalProperties in request body schemas
**Problem description**

The Design Guide does not say how request bodies should handle JSON fields that a client supplies but the schema does not declare. Two sections point in opposite directions:

- §6.1 (Good practices for securing REST APIs) says: *"Validate the request parameters as the first step before they reach the application logic. Implement strong validation checks and immediately reject the request if validation fails."*
- §7 (Compatibility Management) endorses Postel's Law: *"Be conservative in what you do, be liberal in what you accept from others"* — in the immediate context of fields that have been removed from the spec.

Two further pieces of the existing framework matter here:

- §3.2 already defines `400 INVALID_ARGUMENT` with the message *"Request body does not comply with the schema"*. The error response exists; what's missing is guidance on when servers are expected to raise it for unknown fields.
- OAS 3.0.3 default for `additionalProperties` is `true`. Any object schema that does not explicitly set the keyword is implicitly permissive.

Today this combination leaves the policy undefined at the Design Guide level. The OWASP API3:2023 (Broken Object Property Level Authorization) Spectral rules `owasp:api3:2023-no-additionalProperties` and `owasp:api3:2023-constrained-additionalProperties` were added per #539, but both are currently set to `off` in [`.spectral-r4.yaml`](https://github.com/camaraproject/tooling/blob/main/linting/config/.spectral-r4.yaml) pending the resolution of #615.

A practical observation about the rules themselves: their `given` JSONPath matches only object schemas where `additionalProperties` is *explicitly* truthy. Schemas that **omit** the keyword (relying on the OAS 3.0.3 default of `true`) are not matched. I verified this by re-enabling both rules locally and linting QualityOnDemand's three API specs — 0 findings, because QualityOnDemand never sets `additionalProperties` explicitly. So even un-muted, the rules would not nudge the current default-permissive practice; they would only flag schemas where someone wrote `additionalProperties: true` or a non-boolean schema (the `HTTPSettings.headers` case in `CAMARA_event_common.yaml`).

That's the gap I'd like to discuss: the Design Guide doesn't take a position, and the existing rule layer can't substitute for one.

**Possible evolution**

I'd propose adding explicit guidance with this shape (open to refinement):

1. **Default for new CAMARA APIs:** request body schemas SHOULD set `additionalProperties: false`. Unknown fields are rejected with `400 INVALID_ARGUMENT` per §3.2.
2. **Documented exceptions** where permissive `additionalProperties` is justified: each case must (a) cite an external spec it inherits from, or (b) document an explicit API-level extensibility requirement. In both cases, the schema should constrain the size with `maxProperties` to reduce mass-assignment surface.
3. **Reconciling with the §7 Postel reference:** scope the Postel framing to *value tolerance* (e.g., accept multiple date formats) rather than *field-set tolerance*. The existing §7 "tolerant reader" guidance for consumers (clients tolerating server responses) is unaffected.

**Alternative solution**

Document the existing permissive default explicitly — recommend that request bodies leave `additionalProperties` unset, matching the OAS 3.0.3 default — and read §6.1's "strong validation" as applying only to declared parameters. This aligns with §7's existing Postel framing but leaves the OWASP API3:2023 surface unmitigated and means the S-314 / S-316 rules can't be un-muted. Listed here for completeness; I think the strict direction is the right one.

**Additional context**

- **CAMARA's versioning contract does not need permissiveness.** Semver backwards compatibility in CAMARA is "old client + new server": adding optional fields server-side is fine because old clients don't send them. Strict request validation does not interfere with that — strict rejects fields the client should not have sent in the first place against the version they're targeting. The opposite direction (new client + old server) is a version-negotiation question, not a Postel-vs-strict question, and is unchanged here.
- **Transition for existing APIs.** Adding `additionalProperties: false` to an API that previously left it implicit is technically a tightening of validation. Same pattern as the OWASP rules introduced via #539 — start as warnings, tighten over time. Existing APIs would adopt strict at the next MAJOR version bump; new APIs from day one.
- **Current practice across CAMARA** (sampled 10 API repos): roughly a quarter use explicit `additionalProperties: false` on at least some request schemas (CapabilitiesAndRuntimeRestrictions, ApplicationEndpointRegistration, OptimalEdgeDiscovery, SimpleEdgeDiscovery, TrafficInfluence, WebRTC); the rest leave it default-permissive (QualityOnDemand, DeviceLocation, NumberVerification, others). The one explicit permissive case is `HTTPSettings.headers`, which inherits from the CloudEvents HTTP Protocol Binding spec — already being separately worked on in #615.
- **Related closed work:** #549 (per-rule inventory follow-up to #539), #585 (event-subscription-template OWASP alignment).
- **Ask to implementations:** if you operate a CAMARA API in production or pilot, please comment with how your server currently handles unknown JSON fields on request bodies — strict reject, silent ignore, log + accept, other. This will help calibrate whether existing implementations are already aligned with a strict default or whether moving there requires a transition.
- **Mass-assignment background (OWASP API3:2023):** the typical exploit is a client sending an extra field (e.g., `is_admin: true`, `price_override: 0.01`) that a permissive server deserializes onto an internal object whose property of the same name was never meant to be client-controllable. Schema-level `additionalProperties: false` is the first wall, with framework-level allowlists as defence in depth.

hdamker:
#### What type of PR is this?

* correction

Adds explicit Design Guide guidance on how request body schemas should handle JSON properties that a client supplies but the schema does not declare. Today §6.1 ("strong validation checks and immediately reject the request") and §7.4 Postel's Law ("be liberal in what you accept") point in opposite directions, OAS 3.0.3 defaults to permissive, and the existing OWASP `api3:2023` Spectral rules are muted because they cannot enforce a default the Design Guide has not set. #632 captures the gap in detail.

The PR resolves it in the direction of strictness for request body content while preserving Postel's Law for value-level tolerance and tolerant-reader behaviour for responses. Under OAS 3.0.3, `additionalProperties: false` does not propagate through `allOf` / `oneOf` / `anyOf` composition — common in CAMARA APIs via shared types such as `Device` and `Area`. The PR therefore states the rule as a server-side contract: each API declares in `info.description` that the server rejects undeclared properties at any nesting level. `additionalProperties: false` on flat schemas is recommended in addition, where it works.

Seven edits across five sections of `documentation/CAMARA-API-Design-Guide.md`:

1. New §2.3 "Strictness of Request Body Schemas".
2. New §3.2.4 "Request Body Strictness — Mandatory Template for `info.description` in CAMARA API".
3. §5.3.2 index — list §3.2.4 alongside the other REQUIRED `info.description` sections.
4. §5.3.3 example comment — list the new mandatory text.
5. §6.1 item 3 — append sentence cross-referencing §2.3 and §3.2.4.
6. §7.4 Postel's Law bullet — reframe to value-level tolerance.
7. §7.4 "more restrictive validation" bullet — add carve-out + minor-version-bump recommendation.

Fixes #632

- [ ] Yes
- [x] No

The PR is documentation-only. The new rule applies to APIs targeting Commonalities r4.3 and later; APIs targeting earlier Commonalities versions are unaffected.

**Versioning.** For existing APIs adopting strict validation, this PR recommends a minor version bump (PedroDiez 2026-05-14, [issuecomment-4449226832](https://github.com/camaraproject/Commonalities/issues/632#issuecomment-4449226832)). This supersedes the major-version-bump suggestion in the original issue body's "Additional context" section.

**Out of scope by design:**
- HTTP headers (transport layer; CAMARA's permissive header model is intentional).
- Query parameters (different threat model, different OAS mechanism; warrants a sibling PR if/when demand arises).
- `HTTPSettings.headers` ([#615](https://github.com/camaraproject/Commonalities/issues/615)) is an HTTP-header field, not a JSON body field, and is not affected by this PR.
- OAS 3.1 / `unevaluatedProperties` migration ([discussion #144](https://github.com/camaraproject/Commonalities/discussions/144)).
- Un-muting S-314 / S-316 in `.spectral-r4.yaml` — separate tooling PR after this one merges.

**Follow-up work identified during this PR.** The four API/notification sample templates in `Commonalities/artifacts/api-templates/` and `Commonalities/artifacts/notification-templates/` currently do not include the existing §3.2.3 or §6.4 mandatory `info.description` text. Updating those templates to include the new §3.2.4 text alone would leave them inconsistently compliant, so a separate follow-up PR will address all mandatory `info.description` sections (§3.2.3, §3.2.4, §6.4) uniformly across the sample templates.

#### Changelog input

```
release-note
Adds Design Guide guidance on `additionalProperties` for request body schemas: §2.3 (rule), §3.2.4 (mandatory `info.description` template), updates to §6.1 input validation, §7.4 Postel's Law scope, and §7.4 versioning guidance.
```

### DIFF
--- a/documentation/CAMARA-API-Design-Guide.md
+++ b/documentation/CAMARA-API-Design-Guide.md
@@ -11,6 +11,7 @@ This document outlines guidelines for API design within the CAMARA project, appl
 - [2. Data](#2-data)
   - [2.1. Common Data Types](#21-common-data-types)
   - [2.2. Data Definitions](#22-data-definitions)
+  - [2.3. Strictness of Request Body Schemas](#23-strictness-of-request-body-schemas)
 - [3. Responses](#3-responses)
   - [3.1. Business-level Outcomes in Successful Responses](#31-business-level-outcomes-in-successful-responses)
   - [3.2. Error Responses](#32-error-responses)
@@ -219,6 +220,17 @@ When IpAddr is used in a payload, the property `objectType` MUST be present to i
     }
 }
 ```
+
+
+### 2.3. Strictness of Request Body Schemas
+
+CAMARA APIs reject JSON request bodies that contain properties not declared in the API specification. The following rules apply:
+
+- A server MUST reject a request body containing JSON properties not declared in the API specification, at any nesting level. The rejection MUST produce a `400 INVALID_ARGUMENT` response per [3.2. Error Responses](#32-error-responses). The rule is restated to API consumers via the mandatory `info.description` template defined in [Section 3.2.4](#324-request-body-strictness---mandatory-template-for-infodescription-in-camara-api).
+- Where a request body schema and all its referenced schemas are free of composition (no `allOf` / `oneOf` / `anyOf`), the specification SHOULD also declare `additionalProperties: false` on the schema so validators can enforce it.
+- Under OAS 3.0.3, `additionalProperties: false` does not propagate through `allOf` / `oneOf` / `anyOf`. For composed request body schemas — including any schema referencing composed types from `CAMARA_common.yaml` — the mandatory [Section 3.2.4](#324-request-body-strictness---mandatory-template-for-infodescription-in-camara-api) template is the binding rule.
+- The rule applies to CAMARA APIs targeting Commonalities r4.3 and later. APIs targeting earlier Commonalities versions are unaffected.
+- The rule applies to JSON request bodies only. Response-side tolerant-reader behaviour is unchanged (see [Section 7.4](#74-backward-and-forward-compatibility)).
 
 
 
@@ -568,6 +580,18 @@ As a specific rule, error `501 - NOT_IMPLEMENTED` can be only a possible error r
 ```
 
 
+#### 3.2.4. Request Body Strictness - Mandatory Template for `info.description` in CAMARA API
+
+The following template MUST be used as part of the API documentation in the `info.description` property of the CAMARA API specification to declare that the API rejects undeclared JSON properties in request bodies. See [Section 2.3](#23-strictness-of-request-body-schemas) for the underlying rule.
+
+```md
+# Request body strictness
+
+This API rejects requests with JSON request bodies that contain properties not declared in this specification, at any nesting level. Unknown properties result in a `400 INVALID_ARGUMENT` response.
+
+```
+
+
 
 ## 4. Pagination, Sorting and Filtering
 
@@ -834,6 +858,7 @@ info:
   # description explaining the API, part of the API documentation
   # including mandatory text for "Authorization and authentication"
   # including mandatory text for "Additional CAMARA error responses"
+  # including mandatory text for "Request body strictness"
 
 
   description: |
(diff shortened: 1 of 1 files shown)


bigludo7 (approved):
LGTM to us - checked with @patrice-conil

PedroDiez (approved):
LGTM

rartych (approved):
LGTM

PR #637, opened 2026-05-20
docs: minor grammar change for 'identifying the device from the access token"

### LINKED ISSUE #636: docs: minor grammar issue in 'Identifying the device form the access token'
**Problem description**

Three files (info-description-templates.yaml, sample-service.yaml, sample-implicit-events.yaml) include the following sentence:

"This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that _is_ associated with the access token and _was_ identified during the authentication process." _(my italics)_

The mix of case (is, was) may cause confusion to non-native speakers, and can be simplified to:
"[...] the information associated with the access token that was identified during the authentication process."

**Expected action**

Update the three files with the suggested text.

**Additional context**

Kevsy:
#### What type of PR is this?

* documentation

Corrects the grammar for the three files where "identifying the device from the access token" templates are provided

Fixes #636 

- [ ] Yes
- [X] No

#### Changelog input

```
 release-note

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/artifacts/api-templates/sample-implicit-events.yaml
+++ b/artifacts/api-templates/sample-implicit-events.yaml
@@ -39,7 +39,7 @@ info:
     - When the API is invoked using a two-legged access token, the subject will be identified from the optional `device` object, which therefore MUST be provided.
     - When a three-legged access token is used however, this optional identifier MUST NOT be provided, as the subject will be uniquely identified from the access token.
 
-    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information that is associated with the access token and was identified during the authentication process.
+    This approach simplifies API usage for API consumers using a three-legged access token to invoke the API by relying on the information associated with the access token that was identified during the authentication process.
 
     ## Error handling:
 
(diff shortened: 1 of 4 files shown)


PedroDiez:
NOTE:

- It also needs to be updated in [CAMARA-API-Design-Guide.md#appendix-a](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Design-Guide.md#appendix-a-normative-infodescription-template-for-when-user-identification-can-be-from-either-an-access-token-or-explicit-identifier)
- [03/jun update] also needs to be updated in [common/info-description-templates.yaml](https://github.com/camaraproject/Commonalities/blob/main/artifacts/common/info-description-templates.yaml#L109)

Kevsy:
HI @PedroDiez - I think that's all the remaining instances corrected now.

PedroDiez (approved):
LGTM

eric-murray (approved):

bigludo7 (approved):
LGTM

rartych (approved):

hdamker:
@Kevsy @camaraproject/commonalities_codeowners I've created #699 to revert this PR, at least for r4.4 as there is not automatic update of the API definitions in place (and maybe even shouldn't be) - it would create a lot of work for API maintainers without a real value.

Issue #638, opened 2026-05-22
sinkCredential in response schemas contradicts Event Guide §2.2.3


hdamker:
**Problem description**

[Commonalities#633](https://github.com/camaraproject/Commonalities/pull/633) ("Fix issue #619 - Simplify usage of PrivateKeyJwtCredentials for event notification", merged 2026-05-19, included in r4.3) introduces a partial-disclosure model for `sinkCredential` (public client-configuration fields exchanged via request and response; secret fields kept out of responses via `writeOnly: true`). Per the PR description, adding `sinkCredential` to the response shapes is part of the design, not a side effect; the same applies to the matching changes in two sample artifacts (added on review request in the PR thread).

The new model is described in §4.3.1 "Pre-requisites for using credential type `PRIVATE_KEY_JWT`" of the Event Subscription and Notification Guide:

> This information could be shared out of band (manual exchange, contract data, via the Operate API, etc.) or as properties of `sinkCredential` object in subscription request/response.

The earlier (pre-#633) blanket rule in §2.2.3 "Subscriptions Data Model", `sinkCredential` row in the table, was not updated:

> The sinkCredential MUST NOT be present in `POST` and `GET` responses.

The two statements now contradict each other inside the same document. The §2.2.3 wording predates the `writeOnly`-based granular model and reads as stale.

**Sample artifacts changed in the same PR** to align with the new §4.3.1 model:

* `artifacts/api-templates/sample-service-subscriptions.yaml` — added `sinkCredential` to the `Subscription` schema (POST `/subscriptions` 201 response, GET single, GET list).
* `artifacts/api-templates/sample-implicit-events.yaml` — added `sinkCredential` to the `Resource` schema (POST + GET response). The new property sits directly below the existing `sink:` description that reads "`sinkCredential` is deliberately not echoed in responses." — that description was written under the old §2.2.3 rule and is now stale too.

The contradiction was caught when the new validation framework rule `check-sinkcredential-not-in-response` (introduced by [tooling#295](https://github.com/camaraproject/tooling/pull/295)) was run against the r4.3 samples in `camaraproject/ReleaseTest`. Commonalities CI does not run the validation framework against `artifacts/api-templates/`, so the contradiction surfaced only on downstream consumption.

**Expected behavior**

[...]

PR #646, opened 2026-06-11
Define common Sink schema to be used by API templates

### LINKED ISSUE #647: Define common Sink schema to be used by API templates
**Problem description**
The API templates repeat the definition of the sink property. By defining a common schema, this duplication can be avoided.

**Possible evolution**
Define a common Sink schema to be used by the API templates

**Alternative solution**
None - leave things the way they are

**Additional context**
None

eric-murray:
#### What type of PR is this?
* enhancement/feature

The API templates repeat the definition of the `sink` property. By defining a common schema, this duplication can be avoided.

Fixes #647 

- [ ] Yes
- [X] No

None

#### Changelog input

```
 release-note
 - Define common Sink schema to be used by API templates
```

#### Additional documentation 
None

### DIFF
--- a/artifacts/api-templates/sample-implicit-events.yaml
+++ b/artifacts/api-templates/sample-implicit-events.yaml
@@ -442,7 +442,12 @@ components:
         device:
           $ref: "../common/CAMARA_common.yaml#/components/schemas/Device"
         sink:
-          $ref: "../common/CAMARA_event_common.yaml#/components/schemas/Sink"
+          description: |
+            The HTTPS URL where event notifications for this resource will be
+            delivered. Omit this field (and `sinkCredential`) to create the
+            resource without event notifications.
+          allOf:
+            - $ref: "../common/CAMARA_event_common.yaml#/components/schemas/Sink"
         sinkCredential:
           $ref: "../common/CAMARA_event_common.yaml#/components/schemas/SinkCredential"
 
@@ -462,7 +467,11 @@ components:
         device:
           $ref: "../common/CAMARA_common.yaml#/components/schemas/Device"
         sink:
-          $ref: "../common/CAMARA_event_common.yaml#/components/schemas/Sink"
+          description: |
+            The sink URL registered for notifications on this resource, if any.
+            `sinkCredential` is deliberately not echoed in responses.
+          allOf:
+            - $ref: "../common/CAMARA_event_common.yaml#/components/schemas/Sink"
         sinkCredential:
           $ref: "../common/CAMARA_event_common.yaml#/components/schemas/SinkCredential"
 
--- a/artifacts/api-templates/sample-implicit-events.yaml
+++ b/artifacts/api-templates/sample-implicit-events.yaml
@@ -442,15 +442,7 @@ components:
         device:
           $ref: "../common/CAMARA_common.yaml#/components/schemas/Device"
         sink:
-          type: string
-          format: uri
-          maxLength: 2048
-          pattern: ^https:\/\/.+$
-          description: |
-            The HTTPS URL where event notifications for this resource will be
-            delivered. Omit this field (and `sinkCredential`) to create the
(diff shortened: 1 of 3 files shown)


artifacts/api-templates/sample-implicit-events.yaml
```diff
@@ -442,15 +442,7 @@ components:
...
-          type: string
-          format: uri
-          maxLength: 2048
-          pattern: ^https:\/\/.+$
-          description: |
-            The HTTPS URL where event notifications for this resource will be
-            delivered. Omit this field (and `sinkCredential`) to create the
-            resource without event notifications.
-          example: "https://endpoint.example.com/sink"
+          $ref: "../common/CAMARA_event_common.yaml#/components/schemas/Sink"
```
PedroDiez: In order to not miss the specific description we could model (anyway fine to me)

```yaml
sink:
   allOf:
     - description: |-
            The HTTPS URL where event notifications for this resource will be
            delivered. Omit this field (and `sinkCredential`) to create the
            resource without event notifications.
     - $ref: "../common/CAMARA_event_common.yaml#/components/schemas/Sink"
```

WDYT? @eric-murray @patrice-conil, @bigludo7, @rartych

artifacts/api-templates/sample-implicit-events.yaml
```diff
@@ -470,12 +462,7 @@ components:
         device:
           $ref: "../common/CAMARA_common.yaml#/components/schemas/Device"
         sink:
-          type: string
-          format: uri
-          maxLength: 2048
-          description: |
-            The sink URL registered for notifications on this resource, if any.
-            `sinkCredential` is deliberately not echoed in responses.
+          $ref: "../common/CAMARA_event_common.yaml#/components/schemas/Sink"
```
PedroDiez: Idem, in order to not miss the specific description we could model (anyway fine to me)

```yaml
sink:
   allOf:
     - description: |-
            The sink URL registered for notifications on this resource, if any.
            `sinkCredential` is deliberately not echoed in responses.
     - $ref: "../common/CAMARA_event_common.yaml#/components/schemas/Sink"
```

WDYT? @eric-murray @patrice-conil, @bigludo7, @rartych

PedroDiez (commented):
LGTM.

Good point to apply this simplification in the templates it will also help API initiatives for doing alignment work

Minor comment in case we think could be better to consider 
(For the record anyway in case we keep as is, as a way to customize descriptions when needed)

artifacts/api-templates/sample-implicit-events.yaml
eric-murray: Hi @PedroDiez 

I'm happy to keep the local description in the templates, but the usual pattern to override the schema description is the following:
```
sink:
  description: |-
    The HTTPS URL where event notifications for this resource will be
    delivered. Omit this field (and `sinkCredential`) to create the
    resource without event notifications.
  allOf:
    - $ref: "../common/CAMARA_event_common.yaml#/components/schemas/Sink"
```

This works in the Swagger and Redoc viewers. Is there a reason to keep the pattern you propose?

PedroDiez: Hi @eric-murray,

I didn't check that way. It is better approach. Fine going with it

eric-murray:
I updated the proposal to keep the sink description that is currently in the templates

PedroDiez (approved):
LGTM

bigludo7 (approved):

rartych (approved):
LGTM

hdamker:
Heads-up on a merge interaction with #656: that PR updates the `sinkCredential` sentence in the `Resource.sink` description of `sample-implicit-events.yaml` (per #638), while this PR restructures the same property. Whichever merges second will hit a conflict there — the resolution needs to keep both the `allOf` structure from this PR and the updated sentence from #656.

Issue #438, opened 2025-03-31
Adding a section in API `info.description` to provide rationale for not documented Error responses and a reference for their format


PedroDiez:
**Problem description**
In CAMARA MetaRelease Spring25 it has been decided to not document every possible error an API Server may handle (e.g. 5xx errors). As indicated in [API Design Guidelines Section 6.1](https://github.com/camaraproject/Commonalities/blob/main/documentation/API-design-guidelines.md#61-standardized-use-of-camara-error-responses) documentation of non-mandatory error statuses depends on the specific considerations and design of the given API, with additional information provided.

This may be confusing for API developers and implementers, as they fail to see errors in the specification that the implementation can support.

Indeed in case of TEF, some API implementations have had doubts regarding this aspect and the format of these errors. 

Therefore, it is proposed to add a paragraph within the API specification in the `info.description` section that explains this context and provides a reference to these errors so that implementations know how to implement them in case the implementation supports them.

This will help Developers, API implementators and in General CAMARA Community and will bring clarity on this topic.

**Expected action**
Define a template in Commonalities to add a new section in `info.description` field of the API Specification to indicate the context of not documented errors and where their definition can be found.

Proposed template: 
- Avoiding initially the use of links to specific documents as every API version will match to a specific CAMARA Commonalities Release.
- Re-wording may be required due to the refactor of Commonalities API Design Guidelines.  

```md
# Error statuses not documented in API specification
 
Not all error responses are documented in the API as defined by CAMARA. Therefore the API specification may not document some non-mandatory error statuses as indicated in `CAMARA API Design Guidelines`.

For more information please check the CAMARA `API Design Guidelines` document corresponding to the CAMARA Commonalities Release/Version associated with this API version. The format of not documented error statuses supported by the API implementacion can be also checked in the `CAMARA_common.yaml` artifact corresponding to the CAMARA Commonalities Release/Version associated with this API version.

The CAMARA Commonalities Release can be found in the `API Readiness Checklist` document of this API version. The CAMARA Commonalities Version can be found in the `Release Notes` of this API version. 
```

**Additional context**

Kevsy:
HI @PedroDiez -

I think the proposed template is maybe too long and requires the API consumer to do some work to get the information. We could  summarise it:

```
# Additional CAMARA error responses

The list of error codes in this API specification is not exhaustive. Please refer to [CAMARA_common.yaml]({link_to_release}/CAMARA_common.yaml) for a complete list of error responses. 
```

[...]

PR #665, opened 2026-07-06
Minimal referenceable error responses + reusable error examples

### LINKED ISSUE #662: Make `INVALID_TOKEN_CONTEXT` opt-in in the shared `Generic403`
**Problem description**

The shared `Generic403` in `artifacts/common/CAMARA_common.yaml` lists `INVALID_TOKEN_CONTEXT` as an active enum value with a `GENERIC_403_INVALID_TOKEN_CONTEXT` example. Per #369 and the API Design Guide §6.1, this code applies **only** to APIs whose scope is to confirm whether a supplied identity matches the one bound to the three-legged access token (e.g. KYC-Match).

As part of the shared `Generic403`, every API that `$ref`s it inherits the code and declares a `403 INVALID_TOKEN_CONTEXT` it will never legitimately return. Under the 0.8.0 direct-`$ref` model this is now the default for all referencing APIs.

**Expected behavior**

Comment out `INVALID_TOKEN_CONTEXT` in `Generic403` — both the enum entry and the example — making it opt-in, matching the `{{SPECIFIC_CODE}}` precedent already in the same block. Confirm-the-match APIs add the code to their own `Generic403`. Comments are dropped at bundling, so `$ref` resolution and bundled artifacts stay clean.

**Alternative**

Leave the shared `Generic403` as is. Then every API that should not return `INVALID_TOKEN_CONTEXT` must define its own local `Generic403` and cannot `$ref` the common one. That makes the shared `Generic403` unusable for the majority of APIs and pushes a local 403 definition into nearly every spec — the opposite of the direct-`$ref` consolidation that 0.8.0 encourages.

**Additional context**

Scoping background: #259, #369, #321. The Design Guide §6.1 already restricts the code to confirm-the-match scopes; this aligns the shared `Generic403` with that rule.

The same opt-in question applies to other non-universal codes in the shared error responses (e.g. `IDENTIFIER_NOT_FOUND` in `Generic404`, the identifier codes in `Generic422`, `GONE` in `Generic410`); and `CONFLICT` in `Generic409` is marked deprecated but still listed. Those are worth a separate review and are out of scope here.

### LINKED ISSUE #663: Error responses in CAMARA_common.yaml have limited reusability
**Problem description**
As indicated in #662 directly referencing error responses from `CAMARA_common.yaml` can result in incorrect defnitions within API specifications. Initially, the `CAMARA_common.yaml` file was intended to serve as a source for copying text chunks rather than providing fully usable schema definitions for errors.  

Each `Generic4xx` response object in `CAMARA_common.yaml` currently bundles two distinct components:  
- **Schema**: Contains `allOf` with status/code enums. These schemas often need to be API-specific (individual APIs extend the enums for specific codes and exlude irrelevant codes).  
- **Examples**: Provides illustrative static values for each error code. These examples are invariant across all APIs that utilize the corresponding error codes.  

To avoid inconsistencies any response with API-customizable codes can not be `$ref`'ed to `CAMARA_common.yaml` response. 

**Expected behavior**
1. Add a `components/examples` section to `CAMARA_common.yaml` and move all `GENERIC_xxx_CODE` example objects to this section. 
2. API specifications should define their full local response structure based on `ErrorInfo` schema and cherry-pick reusable examples via `$ref`.  API-specific examples may be added locally.
3. The `Generic4xx` response objects can be relocated to API template artifacts for better clarity.

**Alternative solution**
1. Add a `components/examples` section to `CAMARA_common.yaml` and move all `GENERIC_xxx_CODE` example objects to this section  (similar to the expected behavior p.1 and 2).  
2. Split response components based on invariance:  
   - Keep fully-reusable, most common ones in `CAMARA_common.yaml` to be referenced directly.
   - Relocate variable, API-specific response templates to API template artifacts for customization.  

**Additional context**
Current API templates artifacts may need to be updated to follow the choosen solution.

hdamker:
#### What type of PR is this?

* enhancement/feature

Implements the proposed "Option 3" in #663: the `Generic*` error responses bundle error codes that do not apply to every API, so every API referencing them declares error codes it can never return (#662 is the acute case).

- `CAMARA_common.yaml`: adds a catalogue of 10 minimal, directly referenceable error responses (`BadRequest400`, `BadRequestWithRange400`, `Unauthenticated401`, `PermissionDenied403`, `NotFound404`, `IdentifierNotFound404`, `AlreadyExists409`, `DeviceIdentifier422`, `PhoneNumberIdentifier422`, `TooManyRequests429`), each containing only error codes every referencing API can return, plus a new `components/examples` section with the canonical error examples — including those of codes without a referenceable response — for APIs defining error responses locally. Identifier-code examples come in `_DEVICE` / `_PHONE_NUMBER` wording variants; per HTTP status an intro comment lists the available responses and example ingredients.
- `CAMARA_event_common.yaml`: adds `CreateSubscriptionConflict409`, `SinkGone410` (replacing `Generic410` for notification callbacks) and the subject-flavoured `CreateSubscriptionDevice422` / `CreateSubscriptionPhoneNumber422`; existing responses now reference the shared examples, removing the example duplication between the two files.
- The `Generic*` responses and `CreateSubscriptionUnprocessableEntity422` stay in place as deprecated and content-frozen, with two bug fixes: `INVALID_TOKEN_CONTEXT` removed from `Generic403` (#662) and the deprecated 409 code `CONFLICT` removed (enum and example).
- The API templates, the notification template, and Design Guide section 3.2 are updated to the new consumption model: reference a catalogue response, or define the response locally and reference the shared examples.

Fixes #663
Fixes #662

- [ ] Yes
- [x] No

No released API references the `CAMARA_common.yaml` responses via external `$ref` (all released definitions hold local copies), and the deprecated responses remain valid referencing targets. Removal of the deprecated responses is proposed for the next meta-release cycle.

- `sample-notification.yaml` also switched to the new response names and now declares `410` (`SinkGone410`), which the Event Subscription and Notification Guide requires for the notification flow but the template was missing.
- A `GENERIC_400_INVALID_TOKEN` example was added to `CAMARA_event_common.yaml`: the `INVALID_TOKEN` code was listed in the enum of `CreateSubscriptionBadRequest400` but had no example anywhere — please review the proposed wording.
- `GENERIC_404_IDENTIFIER_NOT_FOUND` and `GENERIC_422_SERVICE_NOT_APPLICABLE` now use subject-neutral wording (previously device-worded, also for phone-number APIs); the message example in Design Guide section 3.2.2 is aligned.
- Response names follow the purpose/flavour + status-suffix precedent set by `CAMARA_event_common.yaml`; the `Generic*` names are not reused, to avoid silently changing the meaning of existing references.

#### Changelog input

```
 release-note
Added minimal referenceable error responses and a shared components/examples section to CAMARA_common.yaml and CAMARA_event_common.yaml; Generic* error responses and CreateSubscriptionUnprocessableEntity422 are deprecated (removal planned for the next meta-release cycle); INVALID_TOKEN_CONTEXT removed from Generic403; deprecated 409 code CONFLICT removed.
```

#### Additional documentation

This section can be blank.


artifacts/common/CAMARA_common.yaml
```diff
@@ -365,44 +367,309 @@ components:
...
   responses:
-    #######################################################
-    #######################################################
-    # ERROR RESPONSE SCHEMA TEMPLATE
-    #   - Objective: Make normative error `status` and `code` values
-    #   - Schema Template rationale:
-    #     - The `allOf` in content.application/json.schema allows a combination of both the generic ErrorInfo schema and the specific schema for this error response,
-    #       which validates that `status` and `code` have only the specified values.
-    #       This `allOf` is used without discriminator because it does not imply any hierarchy between the models, just 2 schemas that must be independently validated.
-    #######################################################
```
PedroDiez, line 376: Minor comment. 
Maybe we can keep Lines 376-404 as commented info that may help in case an initiative needs to define API-specific error codes, or a code combination not covered by this catalogue

artifacts/common/CAMARA_common.yaml
```diff
@@ -365,44 +367,309 @@ components:
...
+                      - TOO_MANY_REQUESTS
+          examples:
+            GENERIC_429_TOO_MANY_REQUESTS:
+              $ref: "#/components/examples/GENERIC_429_TOO_MANY_REQUESTS"
+
+    # ─────────────────────────────────────────────────────────────────────────
+    # DEPRECATED error responses
+    #
+    # The Generic* responses below are DEPRECATED as of Commonalities 0.9.0
+    # and are planned for removal in the next meta-release cycle. They bundle
```
PedroDiez: Suggestion (up to WG consideration):

Instead of being removed, would prefer to mention to have it Deprecated but documented (with # as noted text) so having them them as a consultation information but not being directly referenciable.

That is to have some guidance for API designers and API Providers implementing errors not explicitly mentioned in the API Specification. We had some discussion in the past about that point in https://github.com/camaraproject/Commonalities/issues/438 

It works with examples section from pragmatical point of view, but i am afraid this generates confusion again in the future.

artifacts/common/CAMARA_common.yaml
```diff
@@ -365,44 +367,309 @@ components:
...
+    #   Generic404 → NotFound404 and/or IdentifierNotFound404
+    #   Generic409 → AlreadyExists409; CreateSubscriptionConflict409
+    #                (CAMARA_event_common.yaml) for subscription creation;
+    #                local definition from the ABORTED / INCOMPATIBLE_STATE
+    #                examples otherwise
+    #   Generic410 → SinkGone410 (CAMARA_event_common.yaml) in notification
+    #                callbacks; local definition from GENERIC_410_GONE otherwise
+    #   Generic422 → DeviceIdentifier422 or PhoneNumberIdentifier422
+    #   Generic429 → TooManyRequests429
+    #   Generic405, Generic406, Generic412, Generic415, Generic5xx →
```
PedroDiez, line 664: NOTE: 
- From our view 405, 406, 412, 415 and 5xx are single error code cases (not foreseen multiple codes would be allowed in the feature) so it could remain directly referenciable.

Anyway this approach will work

documentation/CAMARA-API-Design-Guide.md
```diff
@@ -501,11 +510,13 @@ The Following table compiles the guidelines to be adopted:
...
 |     4      | Service not applicable for the provided identifier                         |       422        |     SERVICE_NOT_APPLICABLE     | The service is not available for the provided identifier. |
 |     5      | An identifier is not included in the request and the device or phone number identification cannot be derived from the 3-legged access token |       422       |      MISSING_IDENTIFIER      | The device cannot be identified. |
 
+`CAMARA_common.yaml` provides these cases as directly referenceable error responses: `IdentifierNotFound404` (cases 2), and the subject-flavoured `DeviceIdentifier422` / `PhoneNumberIdentifier422` (cases 1, 3, 4, 5 — the phone number flavour omits `UNSUPPORTED_IDENTIFIER`, which does not apply to a single identifier type). The corresponding examples under `components/examples` come in `_DEVICE` and `_PHONE_NUMBER` wording variants for the identifier codes, so that the message wording matches the subject type.
```
PedroDiez:
```suggestion
`CAMARA_common.yaml` provides these cases as directly referenceable error responses: `IdentifierNotFound404` (case 2), and the subject-flavoured `DeviceIdentifier422` / `PhoneNumberIdentifier422` (cases 1, 3, 4, 5 — the phone number flavour omits `UNSUPPORTED_IDENTIFIER`, which does not apply to a single identifier type). The corresponding examples under `components/examples` come in `_DEVICE` and `_PHONE_NUMBER` wording variants for the identifier codes, so that the message wording matches the subject type.
```

artifacts/common/CAMARA_common.yaml
hdamker: Good point. Rather than restore the full commented `ErrorResponseSchema` block inline here — the redesign moved copy-me forms out of `CAMARA_common.yaml` to avoid mixing referenceable definitions with templates — the catalogue intro now points to a complete, copy-ready locally-defined response in `artifacts/api-templates/sample-service.yaml` and calls it out as the skeleton for API-specific codes or code combinations this catalogue does not cover. I have updated the intro comment to make that explicit in commit 82a8f1d.

artifacts/common/CAMARA_common.yaml
hdamker: Agree this is worth keeping as consultation guidance rather than losing it. That removal happens in the next meta-release cycle, not in this PR, so I have softened the banner: the `Generic*` responses are now noted as planned for removal *or downgrade to a commented-out reference* in commit 82a8f1d. The exact mechanic — delete versus keeping them as `#` reference text, as discussed in #438 — is a good decision for the WG to take at that removal step. Flagging it for then.

artifacts/common/CAMARA_common.yaml
hdamker: You are right that there is no superset risk for these single-code statuses. The reason they are examples-only rather than referenceable responses is not the superset — it is the Design Guide's "not documented by default" guidance for 405/406/412/415/5xx. Keeping them out of the referenceable catalogue avoids nudging APIs to document statuses the guide says to omit, while the canonical examples stay available for the rare justified case. Happy to revisit if the WG wants to reconsider the "not documented by default" stance separately.

hdamker:
@PedroDiez I've replied to your three comments and adjusted the PR in the latest commit (kept the Generic* responses as deprecated-but-documented guidance, added the pointer for API-specific codes). Could you check whether that addresses your points and resolve the threads if so? Reviews from the other codeowners are very welcome as well, so that I can react on the comments.

PedroDiez (approved):
LGTM

bigludo7 (approved):
LGTM

rartych (approved):
LGTM

eric-murray (approved):

Issue #668, opened 2026-07-08
Add basic CI to lint the common artifacts (yamllint + Spectral)


hdamker:
**Problem description**

Commonalities has no CI. There is no `.github/workflows/` directory and no GitHub Actions run on pull requests or on pushes to `main`, so nothing checks the common artifacts before they land.

Because every CAMARA API repository consumes these artifacts (`artifacts/common/CAMARA_common.yaml`, `artifacts/common/CAMARA_event_common.yaml`, the API templates, the notification template), an invalid change here has a wide blast radius. A concrete case: the `Date` schema in `CAMARA_common.yaml` was merged with indentation that makes the whole file fail to parse as YAML, and it stayed on `main` undetected. A basic `yamllint` run on the PR would have caught it. (The immediate breakage is corrected in a separate PR.)

**Possible evolution**

Add a lightweight CI workflow, triggered on `pull_request` and `push` to `main`, that lints the artifacts. A feasible minimum:

- **YAML lint** over all `artifacts/**/*.yaml`. This alone catches the class of error above and applies to every artifact, including the component-only `common/*.yaml` files — it is the essential baseline.
- **Spectral (OpenAPI) lint** over the artifacts that are full OpenAPI documents — the API templates under `artifacts/api-templates/` (`sample-service.yaml`, `sample-service-subscriptions.yaml`, `sample-implicit-events.yaml`) declare `openapi:`/`paths:` and lint cleanly. These templates `$ref` a large part of the common files, so linting them with references resolved transitively checks the *referenced* common definitions for OpenAPI consistency. **Worth investigating whether that transitive coverage is sufficient:** today the templates reference only a subset of the common files (for `CAMARA_common.yaml`, ~4/26 schemas and ~9/17 responses; `Date` itself is unreferenced), so template-based resolution does not cover the whole file. The `common/*.yaml` files are themselves component-only (`info` + `components`, no `openapi`/`paths`), so the OpenAPI document rules do not apply to them directly and, if checked standalone, would need a scoped ruleset or exclusion.
- optionally **gherkin lint** over `artifacts/testing/*.feature`.

To stay consistent with what API repositories are validated against, reuse the CAMARA linting configuration from `camaraproject/tooling` (the yamllint config and the Spectral ruleset) rather than maintaining a separate copy.

**Alternative solution**

[...]

PR #669, opened 2026-07-08
fix: correct Date schema indentation in CAMARA_common.yaml

hdamker:
#### What type of PR is this?

* bug

The `Date` schema in `artifacts/common/CAMARA_common.yaml` was added at 5-space / 8-space indentation instead of the 4-space / 6-space used by every sibling schema under `components/schemas`. As a result the whole file fails to parse as YAML, so any tooling or API repository that loads `CAMARA_common.yaml` cannot read it.

This PR re-indents the `Date` schema to 4/6 spaces. It is a whitespace-only change with no effect on the schema content.

Related to #668

- [ ] Yes
- [x] No

Whitespace-only change. Before the fix, `yamllint` reports `syntax error: expected <block end>, but found '<block mapping start>'` at the `Date:` line and `wrong indentation: expected 7 but found 8`; after the fix the file parses and yamllint is clean.

#### Changelog input

```
 release-note
Fixed the indentation of the Date schema in CAMARA_common.yaml so the file parses as valid YAML again.
```

#### Additional documentation

This section can be blank.

### DIFF
--- a/artifacts/common/CAMARA_common.yaml
+++ b/artifacts/common/CAMARA_common.yaml
@@ -92,13 +92,13 @@ components:
       maxLength: 256
       example: "b4333c46-49c0-4f62-80d7-f0ef930f1c46"
 
-     Date:
-        type: string
-        description: The date, using [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6) full-date format "YYYY-MM-DD"
-        format: date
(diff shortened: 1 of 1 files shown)


PedroDiez (approved):
LGTM

eric-murray (approved):

rartych (approved):
LGTM

PR #678, opened 2026-07-17
Normalize artifact line endings and Gherkin formatting

LarryHu0217:
#### What type of PR is this?

* correction
* cleanup

#### What this PR does / why we need it:

Normalizes the common artifacts so their checked-in formatting matches the CAMARA lint configuration:

- adds the proposed root `.gitattributes` policy to keep text files on LF and protect binary formats from conversion
- normalizes the four CRLF artifact files identified in #674 to LF
- fixes the 121 Gherkin indentation findings and 7 trailing-space findings identified in #675

The Gherkin scenarios, steps, examples, and YAML values are unchanged.

#### Which issue(s) this PR fixes:

Fixes #674
Fixes #675

#### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

#### Special notes for reviewers:

This PR depends on #673, which introduces the Artifacts Lint workflow. I validated the combined #673 + this PR state locally with the same `v1-rc` tooling pins:

- yamllint over `artifacts/`: pass
- Spectral over API and notification templates: 0 errors
- gplint over feature templates: 0 errors; 16 expected template warnings remain for #676 to document
- `git diff --ignore-all-space` over all changed artifacts: no content differences
- `git diff --check`: pass

Before this change, gplint reported 144 warnings. The remaining 16 are the 10 duplicate-scenario-name and 6 unused-template-variable warnings described in #676.

#### Changelog input

```
 release-note
- Normalized common artifact line endings to LF and added repository-level Git attributes
- Aligned Gherkin feature template indentation and trailing whitespace with CAMARA lint rules
```

#### Additional documentation

This section can be blank.

```
docs

```

PR #673, opened 2026-07-16
Add Artifacts Lint workflow (yamllint, Spectral, gplint)

hdamker:
#### What type of PR is this?

* enhancement/feature

#### What this PR does / why we need it:

Adds the first CI to this repository — the self-contained `Artifacts Lint` workflow proposed in the [first-step scope comment on #668](https://github.com/camaraproject/Commonalities/issues/668#issuecomment-4990145045). It runs on every pull request, on pushes to `main`, and manually:

- **yamllint** over all YAML files in `artifacts/` — including `Github_templates/`, whose installed copies in API repositories are checked by nothing there. Would have caught the `Date` indentation issue (#669) at PR time.
- **Spectral** over the full-OpenAPI templates (`api-templates/`, `notification-templates/`) with the r4 ruleset; the referenced `common/` definitions are checked transitively via `$ref` resolution.
- **gplint** over the Gherkin feature templates (`testing/`).

All three checks block on error-level findings only; warnings and hints are reported but do not block. Lint configurations and tool versions are taken from `camaraproject/tooling` at the same pinned ref (`v1-rc`) API repositories use for CAMARA Validation, so findings here match the validation toolchain.

Content corrections ship in the same PR, bringing current `main` to zero errors across all three checks:

- kebab-case webhook path placeholder in `sample-notification.yaml` (`/your_webhook_notification_url` → `/your-webhook-notification-url`)
- syntactically valid `externalDocs.url` placeholder in `sample-service.yaml` (`{apiRepository}` → `APIRepositoryName`; the adjacent comment still instructs the replacement), harmonized across all three API templates (PascalCase, like real CAMARA repository names)
- trailing space in `Github_templates/.github/ISSUE_TEMPLATE/config.yml`

`artifacts/linting_rules/README.md` now documents what runs and what to do when a check fails.

#### Which issue(s) this PR fixes:

Part of #668 (first step — the fuller validation outlook remains open there).

#### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

#### Special notes for reviewers:

- Resubmission of #672 from a same-repo branch: GitHub does not execute a workflow file that is newly introduced by a fork PR, so the new check could not run there — on this PR it runs live.
- Warning-level findings (19 Spectral, 144 gplint — the latter mostly mechanical indentation in the feature templates) stay visible in the workflow log but do not block; triaging them fix-or-document is the follow-up noted on #668.
[...]

PR #679, opened 2026-07-17
Align notification template with CAMARA lint rules

### LINKED ISSUE #677: sample-notification.yaml: security scheme description and apiRoot rule mismatch
**Problem description**

Spectral with the CAMARA r4 ruleset reports three findings on `artifacts/notification-templates/sample-notification.yaml` (visible in the Artifacts Lint workflow since #673):

1. Warning `camara-properties-descriptions`: `components.securitySchemes.notificationsBearerAuth` has no `description`.
2. Two hints (`camara-api-root-default`, `camara-api-root-description`): the template's server variable is named `apiRoot`, so the CAMARA apiRoot rules apply — but this server is the API consumer's notification endpoint, not a CAMARA API root. The expected standard text ("API root, defined by the service provider, …") and default (`http://localhost:9091`) would be semantically wrong here.

**Expected behavior**

1. Add a description to `notificationsBearerAuth` (e.g. that the notification sender authenticates via a bearer token agreed during subscription, per the Event Guide).
2. Rename the server variable from `apiRoot` to **`sinkRoot`**, keeping the existing consumer-oriented description. The subscription's `sink` (`CAMARA_event_common.yaml`, `Sink` schema) is the full absolute delivery address (`format: uri`, example `https://endpoint.example.com/sink`); in this document the conventional path `/your-webhook-notification-url` is appended to the server URL, so the variable holds the root portion of that address — `sinkRoot`, mirroring the `apiRoot` naming. The apiRoot rules then no longer apply, and the template no longer suggests the notification endpoint is an API root.
3. Change the variable's default from `https://localhost:8080` to **`https://endpoint.example.com`**, aligning with the `Sink` schema's example (`https://endpoint.example.com/sink`): the notification receiver is the API consumer's server, not a locally runnable mock, and the reserved documentation domain signals "replace with your endpoint".

**Alternative solution**

Keep the `apiRoot` name and document the two hints as expected for this template (expected-warnings documentation issue). Renaming is preferred: the hint noise disappears and the naming is semantically clearer.

**Additional context**

Found while classifying the warnings surfaced by the Artifacts Lint workflow (#668).

A survey of the current API repositories (97 API definition files) shows no naming precedent to align with or conflict against: no API publishes a standalone notification-receiver OpenAPI document — notification endpoints are documented as `callbacks` inside the subscription APIs, addressed by runtime expressions (`{$request.body#/sink}`, legacy `{$request.body#/webhook/notificationUrl}`) without any server variable. 94 of 97 files use `apiRoot` as the single server variable for their own API root (remaining 3: legacy/draft specs). The name chosen here therefore sets the convention for the standalone notification document pattern.

LarryHu0217:
#### What type of PR is this?

* correction

Aligns `sample-notification.yaml` with the notification sink semantics described in #677:

- renames the notification server variable from `apiRoot` to `sinkRoot`
- sets its default to `https://endpoint.example.com`, matching the Sink example
- retains the consumer-oriented notification server description
- documents `notificationsBearerAuth` so the security scheme passes the r4 Spectral description rule

Fixes #677

- [ ] Yes
- [x] No

This PR depends on #673, which introduces the Artifacts Lint workflow. I validated the combined #673 + this PR state locally with the same `v1-rc` tooling pins:

- yamllint over `artifacts/`: pass
- Spectral over API and notification templates: 0 errors, 18 existing warnings, and 0 hints
- targeted Spectral validation for `sample-notification.yaml`: no issue-specific findings
- gplint: 0 errors; its existing formatting warnings are addressed separately by #678
- `git diff --check`: pass

#### Changelog input

```
 release-note
- Model the notification template server as a consumer-provided sink and document bearer authentication
```

#### Additional documentation

This section can be blank.

```
docs

```

### DIFF
--- a/artifacts/notification-templates/sample-notification.yaml
+++ b/artifacts/notification-templates/sample-notification.yaml
@@ -18,11 +18,11 @@ externalDocs:
   url: https://github.com/camaraproject/Commonalities
 
 servers:
-  - url: "{apiRoot}"
+  - url: "{sinkRoot}"
     variables:
-      apiRoot:
-        default: http://localhost:9091
-        description: API root, defined by the service provider, e.g. `api.example.com` or `api.example.com/somepath`
+      sinkRoot:
+        default: https://endpoint.example.com
+        description: Can be any notification server address sent by the client application
 
 tags:
   - name: CAMARA Cloud Event
--- a/artifacts/notification-templates/sample-notification.yaml
+++ b/artifacts/notification-templates/sample-notification.yaml
@@ -21,8 +21,8 @@ servers:
   - url: "{apiRoot}"
     variables:
       apiRoot:
-        default: https://localhost:8080
-        description: Can be any notification server address sent by the client application
+        default: http://localhost:9091
+        description: API root, defined by the service provider, e.g. `api.example.com` or `api.example.com/somepath`
 
 tags:
   - name: CAMARA Cloud Event
@@ -74,3 +74,6 @@ components:
       type: http
       scheme: bearer
       bearerFormat: "{$request.body#/sinkCredential.credentialType}"
+      description: |
+        Bearer token for notification delivery. Token format is determined
+        by `sinkCredential.credentialType` in the subscription request.


hdamker:
@LarryHu0217 thanks for picking this up — and apologies first: the description of #677 was accidentally empty when you started (fixed now), so you could only work from the title. You resolved the findings in a perfectly reasonable way given that.

With the full description in place, the intended resolution for the `apiRoot` part goes in the other direction: the server in this template is the API consumer's notification endpoint — the root part of the subscription's `sink` — not a CAMARA API root. So instead of conforming to the apiRoot rules, the variable gets renamed (`apiRoot` → `sinkRoot`), with default `https://endpoint.example.com` (aligned with the `Sink` schema example in `CAMARA_event_common.yaml`) and the existing consumer-oriented description kept. The apiRoot rules then no longer apply to this file at all.

Your `notificationsBearerAuth` description change fits the issue exactly and can stay as it is.

Would you like to rework the PR along these lines? Happy to review quickly. One heads-up: once #673 merges, `sample-notification.yaml` also changes there (webhook path placeholder becomes kebab-case), so a rebase will be needed — and the new Artifacts Lint check will then run on this PR, too.

hdamker (approved):
LGTM now.

Note for reviewers: the change helps also to avoid that the same rules as for the API's `{apiRoot}` are applied by Validation.

PedroDiez (approved):
LGTM

bigludo7 (approved):
LGTM

rartych (approved):
LGTM

PR #680, opened 2026-07-24
Clarify `sink` behaviour in implicit subscription model

### LINKED ISSUE #628: Implicit-subscription template: document unavailability signalling and clarify implementation-side opt-out
**Problem description**

The implicit-subscription API template ([`sample-implicit-events.yaml` lines 291-312](https://github.com/camaraproject/Commonalities/blob/main/artifacts/api-templates/sample-implicit-events.yaml#L291-L312)) makes `sink` optional on the response `Resource` schema. When a consumer supplies a valid `sink` on the create-resource request, today's only way for an implementation to signal "I created the resource but I'm not delivering events" is to omit `sink` from the 200/201 response body — the resource itself is still created with success status. Two gaps follow:

1. **Implementation-side opt-out is undefined.** The guideline doc ([§2.1, line 60](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md#21-instance-based-implicit-subscription)) says an API project's choice to offer the implicit-subscription pattern is OPTIONAL. It does not address whether a specific implementation of an API that *does* offer the pattern may decline event delivery for a given request.
2. **Unavailability signal is undocumented.** The sink-omission convention is not written down anywhere — neither in the template nor in the guideline doc.

[`CAMARA-API-Design-Guide.md` §3.1.3](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Design-Guide.md#313-recommended-modeling-pattern) "Recommended Modeling Pattern" already provides a fitting framework: the "Nullability/presence rules" approach lets the domain-specific outcome field (here, `sink` echoed in the response) carry the primary signal, with optional `contextCode` / `contextMessage` for explanation. Sink-omission is therefore design-guide-compliant Limited Success on 2xx; it just isn't tied back to §3.1.3 anywhere on the implicit-subscription side.

**Possible evolution**

Working group decision needed: **may an implementation of an implicit-subscription-offering API decline event delivery when the consumer supplies a valid `sink`?** If yes:

1. Document the sink-omission contract explicitly in **both**:
   - The template's `Resource.sink` description ([lines 306-312](https://github.com/camaraproject/Commonalities/blob/main/artifacts/api-templates/sample-implicit-events.yaml#L306-L312)) — state that sink-omission in the response when sink was supplied in the request indicates event delivery is unavailable.
   - The guideline doc ([§2.1](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md#21-instance-based-implicit-subscription)) — cross-reference §3.1.3 of the Design Guide as the rationale.
2. RECOMMEND (not mandate) the optional `contextCode` / `contextMessage` fields per Design Guide §3.1.3 to explain *why* event delivery is unavailable. Non-breaking for clients.

A follow-up PR can land both changes after the WG decision.

**Alternative solution**

Forbid implementation-side opt-out: every implementation of an API offering the implicit-subscription pattern MUST deliver events when a valid `sink` is supplied. Then `sink` becomes required in the response (when present in the request), and unavailability moves to a 4xx or 5xx error path. Cleaner contract, but a step further from current implementations.

**Additional context**

- Related: [#589](https://github.com/camaraproject/Commonalities/issues/589) (Business-Level Outcome 2xx vs 4xx scenarios) — the Design Guide §3.1.3 framework that this issue would consume.
- Related: [#612](https://github.com/camaraproject/Commonalities/pull/612) (the implicit-subscription template that this issue refines).
- Cross-reference filed on QualityOnDemand: [QualityOnDemand#556](https://github.com/camaraproject/QualityOnDemand/issues/556).

PedroDiez:
#### What type of PR is this?

* enhancement/feature

Formalises the behaviour for `sink` handling in the implicit-subscription model: when a consumer provides a sink in the request, the API provider must either deliver events or reject the request with an error — silently creating the resource without notifications is no longer permitted.

**Changes**

`documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md`

- Added a Sink delivery paragraph to §2.1 stating:
  - Implementations SHOULD support event delivery when `sink` is provided.
  - If delivery is not supported, the request MUST be rejected with `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED` and the resource MUST NOT be created.
  - Consumers can retry without `sink` (e.g. fall back to polling).

`artifacts/common/CAMARA_event_common.yaml`
- Added a new example `GENERIC_422_EVENT_NOTIFICATIONS_NOT_SUPPORTED`

`artifacts/api-templates/sample-implicit-events.yaml`
- Updated `POST /resources` description to document the `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED` error path.
- Updated `CreateResource.sink` description to clarify that omitting `sink` skips notifications, and when an API provider doesn't support notification delivery at all when `sink` is signalled within the request, an error is returned and resource is not created.
- Updated `Resource.sink` description to reflect that `sink` is echoed only when delivery is accepted and active.
- Added `EVENT_NOTIFICATIONS_NOT_SUPPORTED` to the 422 response enum and added a corresponding example (referenced to new added `CAMARA_Event_common.yaml` example aforementioned) with an actionable error message

`artifacts/testing/sample-implicit-events-template.feature`
- Added test scenario for the case event notification delivery is implemented by the API provider 
- Added test scenario for the case the API provider doesn't support event notification delivery at all

Fixes #628 

- [x] Yes
- [ ] No

- Not clear behaviour so far so both allowing the resorce creation without notifications delivery as well as generating an error would be acceptable

**NOTE**: Today an implementation may accept sink, create the resource, and simply not send events; nothing forbids it. After this PR that is no longer allowed. Whether to document the new 422 is then each API's own decision: an API that wants to allow implementations without notification delivery adds it — a breaking change per §7.4 of the Design Guide, so a major version — while an API that expects every implementation to deliver events adds nothing and stays as it is.(camaraproject/QualityOnDemand#556 tracks that decision for QoD).

- This PR unifies behaviour and introduces a new Error Code `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED`, so backwards compatibility is not kept

#### Changelog input

```
 clarify `sink` behaviour in implicit subscription model

```

#### Additional documentation 

This section can be blank.

```
docs

```


bigludo7 (dismissed):
LGTM

rartych:
Additional test case in [sample-implicit-events-template.feature](https://github.com/camaraproject/Commonalities/blob/main/artifacts/testing/sample-implicit-events-template.feature) can be considered to check the error response when event delivery is not supported.

artifacts/api-templates/sample-implicit-events.yaml
```diff
@@ -450,6 +462,11 @@ components:
             The HTTPS URL where event notifications for this resource will be
             delivered. Omit this field (and `sinkCredential`) to create the
             resource without event notifications.
+
+            If provided, the API provider SHOULD support event delivery. If
+            notification delivery is not supported, the request is rejected
+            with `422 NOTIFICATIONS_NOT_SUPPORTED` and the resource is NOT
```
rartych: Consistent error naming:
```suggestion
            with `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED` and the resource is NOT
```

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
```diff
@@ -77,6 +77,13 @@ Several types of `sinkCredential` could be available in the future, but for now
 | accessTokenExpireUtc | string date-time | An absolute UTC instant at which the access token shall be considered expired.    | optional    |
 | accessTokenType      | string           | Type of access token - MUST be set to `bearer` for now                            | optional    |
 
+**Sink delivery behaviour**
+
+When `sink` is provided in the POST request, the following rules apply:
+- The API provider SHOULD support event delivery. If notification delivery is not supported (e.g. the implicit subscription feature is not implemented by this provider instance), the request MUST be rejected with `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED` and the resource MUST NOT be created.
```
rartych: Clarification for error condition:

```suggestion
- API providers SHOULD implement event delivery support. If a request includes `sink` and the provider does not support notification delivery (i.e. the implicit subscription feature is not implemented by this provider instance), the request MUST be rejected with 422 EVENT_NOTIFICATIONS_NOT_SUPPORTED and the resource MUST NOT be created.
```

PedroDiez:
> Additional test case in [sample-implicit-events-template.feature](https://github.com/camaraproject/Commonalities/blob/main/artifacts/testing/sample-implicit-events-template.feature) can be considered to check the error response when event delivery is not supported.

Thanks @rartych! Good point

artifacts/api-templates/sample-implicit-events.yaml
```diff
@@ -412,6 +418,12 @@ components:
                 status: 422
                 code: PRIVATE_KEY_JWT_NOT_CONFIGURED
                 message: Private key JWT credential type cannot be used because no configuration was pre-shared.
+            GENERIC_422_NOTIFICATIONS_NOT_SUPPORTED:
```
hdamker: Example key aligned with the code:

```suggestion
            GENERIC_422_EVENT_NOTIFICATIONS_NOT_SUPPORTED:
```

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
hdamker: The attribute `sink` must be optional to not contradict the described behavior and the definition within artifacts (this inconsistency wasn't introduced by the PR, it pre-existed):

```suggestion
| sink           | string | https callback address where the notification must be POST-ed, `format: uri` SHOULD be used to require a string that is compliant with [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).  The [security considerations](#43-notifications-security-considerations) SHOULD be followed.                | optional   |
```

artifacts/api-templates/sample-implicit-events.yaml
```diff
@@ -472,7 +489,9 @@ components:
           $ref: "../common/CAMARA_common.yaml#/components/schemas/Device"
         sink:
           description: |
-            The sink URL registered for notifications on this resource, if any.
+            The sink URL registered for notifications on this resource.
+            Present in the response when `sink` was provided in the creation
+            request and event delivery is supported and active.
```
hdamker: Under the new contract, if delivery isn't supported the resource is not created — so every existing resource whose creation carried `sink` had delivery accepted, and the guide says `sink` MUST be echoed. 

```suggestion
            Present in the response whenever `sink` was provided in the creation request.
```

bigludo7 (dismissed):
LGTM
Thanks @PedroDiez for the reminder

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
```diff
@@ -77,6 +77,13 @@ Several types of `sinkCredential` could be available in the future, but for now
 | accessTokenExpireUtc | string date-time | An absolute UTC instant at which the access token shall be considered expired.    | optional    |
 | accessTokenType      | string           | Type of access token - MUST be set to `bearer` for now                            | optional    |
 
+**Sink delivery behaviour**
+
+When `sink` is provided in the POST request, the following rules apply:
+- API providers SHOULD implement event delivery support. If a request includes `sink` and the API provider does not support notification delivery (i.e. the implicit subscription feature is not implemented by this provider instance), the request MUST be rejected with 422 EVENT_NOTIFICATIONS_NOT_SUPPORTED and the resource MUST NOT be created.
```
hdamker: Vocabulary ... we haven't used "provider instance" yet in the document:

```suggestion
- API providers SHOULD implement event delivery support. If a request includes `sink` and the API provider does not support notification delivery (i.e. the implicit subscription feature is not implemented by this API provider), the request MUST be rejected with 422 EVENT_NOTIFICATIONS_NOT_SUPPORTED and the resource MUST NOT be created.
```

artifacts/testing/sample-implicit-events-template.feature
```diff
@@ -71,3 +71,16 @@ Feature: CAMARA Template Artifact - Test scenarios for sample-implicit-events.ya
     And the response property "$.status" is 400
     And the response property "$.code" is "INVALID_TOKEN" OR "INVALID_ARGUMENT"
     And the response property "$.message" contains a user friendly text
+
+  # Event notifications capability not supported scenario
```
hdamker, line 85: The file inherits # Syntax Error scenarios from line 44, so this 422 currently sits in the syntax group. sample-service-template.feature splits the two — adding the second header keeps that structure:

```suggestion
  # Service Error scenarios

  # Event notifications capability not supported scenario
```

artifacts/testing/sample-implicit-events-template.feature
hdamker, line 48: §2.1 now also states that `sink` MUST be echoed in the response when event delivery is accepted, and no scenario covers that yet — this one verifies the notification arrives, and its `When` acts on an already-created resource. A compact scenario for the creation response could look like:

```
  @{feature_identifier}_{operationId}_xx_sink_echoed
  Scenario: Sink is returned in the creation response when event delivery is accepted
    Given the request body includes property "$.sink" with a valid value
    When the request "{operationId}" is sent
    Then the response status code is 2<xx>
    And the response property "$.sink" has same value as the request body property "$.sink"
```

hdamker (commented):
See a few suggestions below.

Two small things in the PR description, since it feeds the release notes:

- The **Changes** list doesn't mention the added test scenario in `artifacts/testing/sample-implicit-events-template.feature`.
- *"providing an unsupported sink returns an error"* (in the `CreateResource.sink` bullet) sounds like the existing `400 INVALID_SINK` / `INVALID_PROTOCOL` / `INVALID_CREDENTIAL` cases, which this PR doesn't change. The new code is for a different case: the provider doesn't support notification delivery at all.

On the breaking change I agree, and I think the PR should say who has to act on it. Today an implementation may accept `sink`, create the resource, and simply not send events; nothing forbids it. After this PR that is no longer allowed. Whether to document the new `422` is then each API's own decision: an API that wants to allow implementations without notification delivery adds it — a breaking change per §7.4 of the Design Guide, so a major version — while an API that expects every implementation to deliver events adds nothing and stays as it is (camaraproject/QualityOnDemand#556 tracks that decision for QoD). Worth saying so in §2.1, so the new text isn't read as requiring the error in every API offering the pattern.

artifacts/api-templates/sample-implicit-events.yaml
PedroDiez: This part us updated from the merge of #665.

I accomodate this suggestion into it

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
PedroDiez: I will also update prior context (*), as `sink` was stated as mandatory in the context that is required to enable implicit subscriptions mode

Line 62:
If this capability is present in CAMARA API, the following attributes MUST be used in the POST request :

PedroDiez: Will be adding this note:

**NOTE**: To request the Implicit Subscription mode, `sink` attribute MUST be provided within the API request body.

artifacts/api-templates/sample-implicit-events.yaml
PedroDiez: done in: [updates_in_events_guide_and_implict_events_template](https://github.com/camaraproject/Commonalities/pull/680/commits/d5aa5920e3255022e28d6f3b3cf40009cf1d127f)

artifacts/testing/sample-implicit-events-template.feature
PedroDiez: Good idea, adding this test to the suite

PedroDiez: Done in [add_test_scenario](https://github.com/camaraproject/Commonalities/pull/680/commits/b03468debdb4203ebc5990be36c8ab0ceca52825)

PedroDiez:
> See a few suggestions below.
> 
> Two small things in the PR description, since it feeds the release notes:
> 
> * The **Changes** list doesn't mention the added test scenario in `artifacts/testing/sample-implicit-events-template.feature`.
> * _"providing an unsupported sink returns an error"_ (in the `CreateResource.sink` bullet) sounds like the existing `400 INVALID_SINK` / `INVALID_PROTOCOL` / `INVALID_CREDENTIAL` cases, which this PR doesn't change. The new code is for a different case: the provider doesn't support notification delivery at all.
> 
> On the breaking change I agree, and I think the PR should say who has to act on it. Today an implementation may accept `sink`, create the resource, and simply not send events; nothing forbids it. After this PR that is no longer allowed. Whether to document the new `422` is then each API's own decision: an API that wants to allow implementations without notification delivery adds it — a breaking change per §7.4 of the Design Guide, so a major version — while an API that expects every implementation to deliver events adds nothing and stays as it is ([camaraproject/QualityOnDemand#556](https://github.com/camaraproject/QualityOnDemand/issues/556) tracks that decision for QoD). Worth saying so in §2.1, so the new text isn't read as requiring the error in every API offering the pattern.

Hi @hdamker, please take a look to the updates

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
```diff
@@ -80,6 +82,19 @@ Several types of `sinkCredential` could be available in the future, but for now
...
+- When event delivery is accepted, `sink` MUST be echoed in the response.
+- The implementation MUST NOT silently create the resource while ignoring the requested `sink`. If event delivery is unavailable, the consumer MUST be informed via the error response, so they can retry without `sink` (e.g. fall back to a polling pattern via `GET /<resource>`).
+
+Aforemention rules imply a breaking change, so as it is worth to state below since Metarelease Sync26:
+- New APIs that includes in their design the implicit subscription model MUST follow aforementioned behaviour.
+- Previous existing APIs that owns in their design the implicit subscription model. Whether to document/align with this behaviour is an each API's own decision:
+  - An API that wants to allow implementations without notification delivery adds it — that implies breaking change per section [7.4. Backward and Forward Compatibility of the CAMARA API Design Guide](./CAMARA-API-Design-Guide.md##74-backward-and-forward-compatibility) —, so a major version is required
+  - An API that expects every implementation to deliver events adds nothing and stays as it is.
```
hdamker: Anchoring on the Commonalities release rather than the meta-release label, following the precedent in [Section 2.3. Strictness of Request Body Schemas](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Design-Guide.md#23-strictness-of-request-body-schemas) of the Design Guide ("The rule applies to CAMARA APIs targeting Commonalities r4.3 and later. APIs targeting earlier Commonalities versions are unaffected."). Also fixes the link to Section 7.4, which currently has a double `##` and so does not resolve, and makes explicit what an API that adds nothing expects from its implementations.

```suggestion
These rules apply to CAMARA APIs targeting Commonalities r4.4 and later. APIs targeting earlier Commonalities versions are unaffected.

- An API newly adopting the implicit subscription model MUST follow the behaviour described above.
- For an API that already offers the implicit subscription model, whether to document `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED` is that API's own decision:
  - An API that allows implementations without notification delivery adds the error code to its API definition. This is a breaking change per [Section 7.4. Backward and Forward Compatibility](/documentation/CAMARA-API-Design-Guide.md#74-backward-and-forward-compatibility) of the CAMARA API Design Guide, so a new major version is required.
  - An API that expects every implementation to deliver event notifications adds nothing and stays as it is. Implementations of such an API MUST deliver event notifications when `sink` is provided.
```

hdamker (commented):
Thanks for the updates, the suggestions are all in. Four small ones below, plus one point that is not code.

`GENERIC_422_EVENT_NOTIFICATIONS_NOT_SUPPORTED` is the only example in `CreateResourceUnprocessableEntity422` defined inline; the other five reference the shared examples. Under the consumption model introduced by #665 (define the response locally, reference the shared examples) the new example would be added to `components/examples` in `CAMARA_event_common.yaml` and referenced from here, as `GENERIC_422_PRIVATE_KEY_JWT_NOT_CONFIGURED` is. That also gives APIs adopting the implicit subscription pattern something to reference instead of copying the example. The intro comment of the 422 block in that section is scoped to subscription creation today, so it would need to cover the implicit model as well.

documentation/CAMARA-API-Event-Subscription-and-Notification-Guide.md
```diff
@@ -59,13 +59,15 @@ The subscription terminates with the managed entity.
 
 Providing this capability is OPTIONAL for any CAMARA API depending on UC requirements.
 
-If this capability is present in CAMARA API, the following attributes MUST be used in the POST request :
+If this capability is present in CAMARA API, the following attributes MUST be used in the POST request:
```
hdamker: With `sink` now optional, both rows of the table are optional, so "attributes MUST be used" reads as contradicting the cardinality column. Mandating the attribute *names* keeps the original intent:

```suggestion
If this capability is present in CAMARA API, the following attribute names MUST be used in the POST request:
```

artifacts/api-templates/sample-implicit-events.yaml
```diff
@@ -408,6 +420,11 @@ components:
             The HTTPS URL where event notifications for this resource will be
             delivered. Omit this field (and `sinkCredential`) to create the
             resource without event notifications.
+
+            If provided, the API provider SHOULD support event delivery. If
+            notification delivery is not supported, the request is rejected
+            with `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED` and the resource is NOT
+            created.
```
hdamker: The operation description states this as `MUST be rejected` / `MUST NOT be created`; the same wording here keeps the two consistent:

```suggestion
            notification delivery is not supported, the request MUST be rejected
            with `422 EVENT_NOTIFICATIONS_NOT_SUPPORTED` and the resource MUST NOT
            be created.
```

artifacts/testing/sample-implicit-events-template.feature
```diff
@@ -71,3 +79,18 @@ Feature: CAMARA Template Artifact - Test scenarios for sample-implicit-events.ya
     And the response property "$.status" is 400
     And the response property "$.code" is "INVALID_TOKEN" OR "INVALID_ARGUMENT"
     And the response property "$.message" contains a user friendly text
+
+  # Service Error scenarios
+
+  # Event notifications capability not supported scenario
```
hdamker, line 85: Since documenting the error is each API's own decision, marking the scenario's applicability follows the convention in `event-subscription-template.feature`, which flags its equivalent case with "Specific Subscription error scenario if multi-event is not permitted":

```suggestion
  # Service Error scenarios

  # Event notifications capability not supported scenario
  # Only applicable if the API defines the error 422 EVENT_NOTIFICATIONS_NOT_SUPPORTED
```

hdamker (approved):
LGTM - thanks for also adding the example to CAMARA_event_common.yaml

(corrected the filename in the PR description, and deleted an invisible U+200E formatting character in another quoted path)

rartych (approved):
LGTM

Thank you @PedroDiez

bigludo7 (approved):
LGTM - Thanks Pedro

PedroDiez:
@eric-murray, please can you review this PR?

eric-murray (approved):

PR #683, opened 2026-07-29
Fix remaining gplint findings in the feature templates

### LINKED ISSUE #682: Feature file templates: fix the remaining gplint findings (placeholder syntax, scenario names)
**Problem description**

After #678, gplint over `artifacts/testing/*.feature` reports 16 warnings: 10× `no-dupe-scenario-names` and 6× `no-unused-variables`. #676 plans to document them as expected; all three groups below are better fixed, and none requires a semantic change.

**1. `sample-service-template.feature` mixes two placeholder conventions (6× `no-unused-variables`)**

The tags and the error scenarios use `{createResource}` / `{operationId}`; three happy-path Scenario Outlines use `<createResource>` / `<Resource>`. Angle brackets are Gherkin's Scenario Outline variable syntax, so gplint reports them as variables absent from the Examples table — and a Gherkin runner would try to substitute them from it. Visible inside one file: line 50 `When the request "<createResource>" is sent` against line 199 `When the request "{operationId}" is sent`. The same hazard sits on `<ResourceId>`, `<listResources>`, `<getResource>`, `<deleteResource>`, `<operationId>`, `<requestProperty>`, `<requestSchema>` and `<scope>`, unflagged only because those scenarios have no Examples table.

**2. `event-subscription-template.feature` reuses scenario names across additive scenarios (5× `no-dupe-scenario-names`)**

Scenarios 12 to 14 (lines 150, 168, 181) carry scenario 01's name "Create \<xxx\> subscription (sync creation)", and scenarios 15 and 16 (lines 199, 212) carry scenario 06's "Get a subscription based on existing subscription-id.". These are additive, not alternatives: a subscription API keeps all of them, so the duplicate names propagate into every `*-subscriptions.feature` copied from the template. The tags already carry the distinction, and downstream APIs already name these scenarios apart (QualityOnDemand, CustomerInsights).

**3. `sample-service-template.feature` reuses one canonical scenario name (5× `no-dupe-scenario-names`)**

Lines 67, 90, 112, 136 and 149 repeat line 44's "Common validations for any success scenario" in two different situations: lines 44/67/90 are three mutually exclusive variants of the same scenario (device-identified API / phoneNumber-identified API / neither), of which an API author keeps one; lines 112/136/149 apply the name to different operations.

**Expected behavior**

1. Normalise every template placeholder in `sample-service-template.feature` to the `{…}` convention, keeping only the genuine Examples variables `<property>`, `<condition>` and `<required_property>`.
2. Give scenarios 12 to 16 in `event-subscription-template.feature` names that carry the distinction their tags already make.
3. Add a distinguishing suffix to the duplicated names in `sample-service-template.feature` — the variant condition for lines 44/67/90, the operation for lines 112/136/149.

No scenario, step, examples table or tag is added or removed. With all three applied, `gplint --config .gplintrc artifacts/testing/*.feature` reports 0 problems and #676 no longer needs a gplint section.

**Alternative solution**

Document all 16 as expected, the current plan in #676. Rejected: group 1 is a defect that every API repository inherits when it copies the template, and group 2 puts duplicate scenario names into every subscription API's test report.

**Additional context**

Best done on top of #678, which rewrites all 550 lines of `event-subscription-template.feature` for line endings — a change developed from `main` instead will conflict there.

"Common validations for any success scenario" appears in 39 feature files across the API repositories. The suffixes in group 3 extend that name rather than replace it, so the convention stays recognisable.

hdamker:
#### What type of PR is this?

* correction

gplint over `artifacts/testing/*.feature` reported 16 warnings. This PR clears all 16 without adding or removing a single scenario, step, Examples table or tag.

**1. Placeholder syntax (6 warnings, plus consistency)**

Angle brackets are Gherkin's Scenario Outline variable syntax. In `sample-service-template.feature` they were also used for hand-substituted placeholders, in scenarios that *do* have an Examples table — so gplint reported them as variables missing from that table, and a Gherkin runner would try to substitute them. The two conventions sat side by side in the same file — `When the request "<createResource>" is sent` in the happy-path scenarios against `When the request "{operationId}" is sent` in the error scenarios. That is the change this PR needs to make.

The same mixed usage sits in `sample-implicit-events-template.feature` and `event-subscription-template.feature`. It reports nothing today only because those placeholders are in plain scenarios with no Examples table to collide with — the trap springs the moment anyone adds one. Leaving them would also mean two conventions across the templates an API repository copies together, so the same normalisation is applied there for consistency.

The result is a rule that is now checkable: **every remaining angle-bracket token in `artifacts/testing/` is a declared Examples column**, so angle brackets carry exactly one meaning. The genuine Examples variables keep them — `<property>`, `<condition>`, `<required_property>`, `<device_identifier>`, `<oas_spec_schema>`, `<unsupported_credential_type>`.

**2. Duplicate scenario names (10 warnings)**

In `event-subscription-template.feature`, scenarios 12 to 16 carried the names of scenarios 01 and 06. These are additive, not alternatives — a subscription API keeps all of them, so the duplicates propagate into every `*-subscriptions.feature` copied from the template. They now carry the distinction their tags already make (sinkCredential type, provisioning mode).

In `sample-service-template.feature`, "Common validations for any success scenario" appeared six times: three mutually exclusive variants of one scenario, of which an author keeps one, and three different operations. Each now carries a suffix — the variant condition, or the operation. The name is extended rather than replaced, matching how API repositories already distinguish it (for example "… with 3-legs authentication"), so it stays recognisable across the 39 feature files that use it.

Verification: gplint with the tooling-pinned config reports 0 problems. Every commit is line-for-line — the counts of `Scenario`, `Scenario Outline`, `Examples:`, tag lines and table rows are identical before and after.

Fixes #682

- [ ] Yes
- [x] No

Built on top of #678, which normalises line endings in the same files — please merge that one first. Developing from `main` instead would have conflicted with it across all 550 lines of the event subscription template.

Draft while #678 is pending: because this branch is based on #678's head, the diff below currently also contains its commit `05d7158`. Once #678 merges I will rebase onto `main`, and the diff reduces to the three commits described here — already verified by replaying them onto a simulated post-merge `main`.

The three commits are separate on purpose: the required fix, then each consistency file, so they can be reviewed independently.

`event-subscription-template.feature` is the largest diff but the smallest change: one placeholder, `<xxx>`, appears 146 times. Its header already described these as substitution placeholders and now reads "We use `{xxx}` as the subscription resource prefix".

#### Changelog input

```
 release-note
Feature file templates: placeholders use the {…} convention consistently, and duplicate scenario names are made distinct. No scenario, step, Examples table or tag is added or removed.
```

#### Additional documentation

This section can be blank.

```
docs

```

### DIFF
--- a/artifacts/testing/event-subscription-template.feature
+++ b/artifacts/testing/event-subscription-template.feature
@@ -147,7 +147,7 @@ Feature: Camara Template Subscriptions API - Operations on subscriptions
 ######################### Additional Happy Path Scenarios ##############################
 
   @<xxx>_subscriptions_12_Create_<xxx>_subscription_sync_with_accesstoken_sink_credential
-  Scenario: Create <xxx> subscription (sync creation)
+  Scenario: Create <xxx> subscription (sync creation) with ACCESSTOKEN sinkCredential
   # Some implementations may only support asynchronous subscription creation
   # Some implementations may decide to not return the sinkCredential in the response (data minimization principle)
     Given that subscriptions are created synchronously
@@ -165,7 +165,7 @@ Feature: Camara Template Subscriptions API - Operations on subscriptions
     And the response body property "$.sinkCredential.accessTokenExpiresUtc", if present, is set to the same value of the request property "$.sinkCredential.accessTokenExpiresUtc"
 
   @<xxx>_subscriptions_13_Create_<xxx>_subscription_sync_with_private_jwt_key_sink_credential_out_of_band_provisioning
-  Scenario: Create <xxx> subscription (sync creation)
+  Scenario: Create <xxx> subscription (sync creation) with PRIVATE_JWT_KEY sinkCredential, out-of-band provisioning
   # Some implementations may only support asynchronous subscription creation
   # Some implementations may only support out_of_band provisioning
     Given that subscriptions are created synchronously
@@ -178,7 +178,7 @@ Feature: Camara Template Subscriptions API - Operations on subscriptions
     And the response body complies with the OAS schema at "#/components/schemas/Subscription"
 
   @<xxx>_subscriptions_14_Create_<xxx>_subscription_sync_with_private_jwt_key_sink_credential_in_band_provisioning
-  Scenario: Create <xxx> subscription (sync creation)
+  Scenario: Create <xxx> subscription (sync creation) with PRIVATE_JWT_KEY sinkCredential, in-band provisioning
   # Some implementations may only support asynchronous subscription creation
   # Some implementations may additionally support in_band provisioning
     Given that subscriptions are created synchronously
@@ -196,7 +196,7 @@ Feature: Camara Template Subscriptions API - Operations on subscriptions
 
   @<xxx>_subscriptions_15_Operation_to_retrieve_subscription_based_on_an_existing_subscription-id_access_token_sink_credential_returned
   # Some implementations may decide to not return the sinkCredential in the response (data minimization principle)
-  Scenario: Get a subscription based on existing subscription-id.
+  Scenario: Get a subscription based on existing subscription-id, with ACCESSTOKEN sinkCredential returned.
     Given the path parameter "subscriptionId" is set to the identifier of an existing <xxx> subscription
     When the request "retrieve<xxx>Subscription" is sent
     Then the response code is 200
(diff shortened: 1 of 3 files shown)


PedroDiez (dismissed):
LGTM

PedroDiez (approved):
LGTM

rartych (approved):
LGTM

bigludo7 (approved):
LGTM

eric-murray (approved):

PR #685, opened 2026-07-29
Artifacts Lint: attribute Spectral findings per template and document the expected baseline

### LINKED ISSUE #676: Document the expected lint warnings of the artifacts (fix-or-document standard)
**Problem description**

API repositories are asked to fix all validation warnings or document why not. The common artifacts should meet the same standard. After the mechanical cleanups (companion issues), a set of warnings remains that is expected rather than fixable, and nothing currently tells a PR author which warnings in the Artifacts Lint output (#673) are the known baseline:

- 9× Spectral `owasp:api4:2023-string-restricted` — free-text/identifier string fields with `maxLength` but no sensible `format`/`pattern`/`enum`: 5 in the common files (`ErrorInfo.code`, `ErrorInfo.message`, `NetworkAccessIdentifier`, `CloudEvent.id`, `CloudEvent.type`; already recorded as known-unactionable in the CAMARA Validation rule metadata, S-313) and 4 in the API templates (`CreateResource.name`, `Resource.name` in two templates; same rationale).
- 9× Spectral `oas3-unused-component` — all nine components are referenced via `discriminator.mapping` entries, which Spectral does not follow (`EventApiSpecific1/2`, `EventSubscriptionStarted/Updated/Ended`, `HTTPSubscriptionRequest/Response`); see camaraproject/tooling#377 for a detection improvement.
- 10× gplint `no-dupe-scenario-names`, 6× `no-unused-variables` — template-natured: scenario variants sharing names across template sections, and placeholder variables intentionally not bound in Examples tables.

**Expected action**

Add an "expected warnings" section to `artifacts/linting_rules/README.md`: one line per group with the rationale, so contributors can distinguish the known baseline from new warnings their change introduces. Update the section when a cleanup or rule change shrinks the baseline.

**Additional context**

The scope of the OWASP string rules is questioned more broadly in #596; the documentation here records the current state and follows whatever that discussion resolves.

### LINKED ISSUE #684: Artifacts Lint: Spectral step loses finding attribution and under-reports the real findings
**Problem description**

The Spectral step of the Artifacts Lint workflow (#673) lints all four templates in a single Spectral invocation. Spectral loses the source file of most findings when several input documents share `$ref` targets ([stoplightio/spectral#2640](https://github.com/stoplightio/spectral/issues/2640)), so the step reports the same result three different ways and spends most of its annotation budget on findings it cannot attribute.

Run 30385534061 (job 90363770115), on the head of #678:

| Surface | Count |
| --- | --- |
| pretty summary on stdout | `✖ 9 problems (0 errors, 9 warnings, 0 infos, 0 hints)` |
| `::warning` lines in the job log | 152 |
| annotations on the check run | 20 — 10 from the Spectral step, 10 from the gplint step |

GitHub renders at most 10 annotations per level per step. 143 of the 152 findings carry no source, so they annotate `.` at line 1: three of the Spectral step's ten slots go to those unattributed annotations, and the two real findings in `api-templates/sample-service.yaml` (lines 278 and 293) get no annotation at all. A contributor reading one step sees three numbers — 9, 20, 152 — for one result.

Reproduced against `main` with the pinned toolchain (Spectral 6.16.2, `.spectral-r4.yaml`); the counts are identical to the run above, so nothing here is specific to #678:

- one invocation over all four templates → 152 results, 9 with a source, 143 with none (`range` 0:0)
- one invocation per template → 26 results, **none** without a source, **15 unique** `(file, line, rule)`

| Template linted alone | Findings | Without source |
| --- | --: | --: |
| `api-templates/sample-implicit-events.yaml` | 7 | 0 |
| `api-templates/sample-service-subscriptions.yaml` | 10 | 0 |
| `api-templates/sample-service.yaml` | 5 | 0 |
| `notification-templates/sample-notification.yaml` | 4 | 0 |

The behaviour is deterministic and independent of argument order, and adding documents shrinks the attributed set (two templates → 12 sourced, four → 9).

The pretty summary is therefore not a reference surface — its 9 findings are a lossy subset of the real 15, and six genuine findings are invisible in the current output entirely:

- `common/CAMARA_common.yaml:57` — the pagination `Link` header
- `common/CAMARA_event_common.yaml` lines 111, 173, 393, 416, 441

**Expected behavior**

Lint each template in its own invocation and deduplicate the results on `(file, line, rule)`. CAMARA Validation already does exactly this for the same reason — `run_spectral_engine` in the tooling repository invokes Spectral per file and cites the same Spectral issue — so the workflow converges with the validation toolchain instead of diverging from it. The pretty output, the annotations and the reported counts then all describe the same 15 findings.

Ten of those 15 are the documented common-library baseline, already recorded in the `suppress_schema_paths` allowlist of rule S-313 in the tooling rule metadata. Bare Spectral does not read that metadata, so the step should apply the allowlist itself: collapse the findings it covers into a single `notice` annotation per source file, carrying that file's count and pointing at the documented list, and annotate everything else individually at its native Spectral severity.

The cap is 10 annotations per level per step, so collapsing the baseline to `notice` takes it off the warning budget entirely instead of merely shrinking its share of it — all 10 warning slots stay available for findings a change actually introduces, against 10 of 10 consumed by the baseline today. Keying the collapse on the allowlisted locations rather than on the rule code preserves the property that allowlist is built for: a newly added unconstrained string in a common schema still surfaces on its own and can be judged on its own merits.

On the current baseline that is 2 `notice` annotations (`common/CAMARA_common.yaml`, `common/CAMARA_event_common.yaml`) and 5 individual `warning` annotations.

**Alternative solution**

- Filter out the source-less lines. Makes the annotations honest but cements the lossy 9-finding view, so the six findings listed above stay invisible.
- Lint a bundled specification instead. Attributes findings to a generated file rather than to `artifacts/common/**`; the tooling documentation describes bundles as a preview and not a means of locating the source of a validation problem.
- Drop `-f github-actions`. Gives up the annotations #673 added and leaves the under-reporting unfixed.

**Additional context**

#676 documents the expected findings of the artifacts. Its recorded baseline of 9 Spectral `owasp:api4:2023-string-restricted` findings is the buggy multi-document view; the real baseline is 15 — 11 in the common files and 4 `name` fields in the API templates. The classification in #676 is unaffected, only the count.

hdamker:
#### What type of PR is this?

* bug

The Spectral step of the Artifacts Lint workflow (#673) linted all four templates in one invocation. Spectral loses the source file of most findings when several input documents share `$ref` targets ([stoplightio/spectral#2640](https://github.com/stoplightio/spectral/issues/2640)), so the step reported 9 problems in its summary, emitted 152 `::warning` lines, and produced 10 annotations of which 3 pointed at `.` line 1 — while the two real findings in `api-templates/sample-service.yaml` had no annotation at all. Details and the reproduction in #684.

Spectral now runs **once per template** and the results are merged, which is what CAMARA Validation already does for the same reason. Six findings that the old view hid become visible, and the log, the annotations and the summary now describe one set of 15 findings instead of three different numbers:

- `.github/scripts/spectral_annotations.py` merges the per-template JSON, deduplicates on `(file, line, rule)`, emits the annotations, and prints a reconciliation table to the log and to the run summary.
- The 10 findings that are the documented common-library baseline are collapsed into one `notice` annotation per source file, pointing at `artifacts/linting_rules/README.md`. The remaining 5 keep an individual annotation at their native severity.
- The allowlist behind that split is read from `validation/rules/spectral-rules.yaml` in `camaraproject/tooling` (rule S-313 `suppress_schema_paths`) — added to the existing sparse-checkout rather than copied, so it cannot drift from what API repositories apply. Matching replicates the tooling semantics, including the dot-boundary prefix rule, so a newly added unconstrained string in a common schema still surfaces on its own.

Annotation budget, per step: **5 of 10 warning slots** used, plus 2 notices — against 10 of 10 consumed by the baseline before this change, where a warning introduced by a pull request could go unannotated.

`artifacts/linting_rules/README.md` gains the expected-findings section from #676: all 15 findings with their rationale, keyed on schema path rather than line number so it does not go stale, and marked collapsed or individual so it can be checked against what the workflow actually annotates.

Fixes #684
Fixes #676

- [ ] Yes
- [x] No

- The two halves are mutually dependent: the aggregate annotation points at the README section, and the section's baseline is the corrected 15-finding set — so they are reviewable only together.
- All numbers were reproduced against `main` with the pinned toolchain (Spectral 6.16.2, `.spectral-r4.yaml`): 152 raw results with 9 sourced for one invocation over four templates, versus 26 raw results with none unsourced and 15 unique for one invocation per template.
- Bare Spectral reads no rule metadata, so the 15-finding baseline is irreducible here: [camaraproject/tooling#401](https://github.com/camaraproject/tooling/issues/401) will not change what this CI shows. Under this change, however, the pagination `Link` header finding stops being annotated individually the moment camaraproject/tooling#401 lands, with no edit needed in this repository.
- Two groups from #676 need no documentation: the 9 `oas3-unused-component` findings are gone since tooling made unused-component detection discriminator-aware, and the gplint findings are being fixed (#682, via #683) rather than documented as expected. The section therefore covers Spectral only.
- Verified on a dispatched run before this PR ([hdamker/Commonalities run 30483686070](https://github.com/hdamker/Commonalities/actions/runs/30483686070)): the Spectral step produces 5 warning annotations at the five locations listed in the README section plus 2 `notice` aggregates, and no annotation points at `.` line 1.
- A Spectral runtime error on one template no longer hides the other three: the step reports what the rest found, annotates the failed template, and then fails.

#### Changelog input

```
 release-note
- Artifacts Lint: Spectral runs per template so findings keep their source file; annotations now cover the findings that need review, the documented baseline is collapsed into one notice per file, and the log and run summary report one reconciled count
- Documented the expected lint findings of the artifacts in artifacts/linting_rules/README.md
```

#### Additional documentation

```
docs
artifacts/linting_rules/README.md — new "Expected lint findings" section
```

### DIFF
--- a/.github/scripts/spectral_annotations.py
+++ b/.github/scripts/spectral_annotations.py
@@ -1,0 +1,424 @@
+#!/usr/bin/env python3
+"""Turn per-file Spectral JSON output into check-run annotations and a report.
+
+Used by .github/workflows/artifacts-lint.yml.  The Spectral step lints each
+template in its own invocation (see the workflow header for why) and drops one
+JSON result file per template into a directory; this script merges them into
+the single view a reviewer should see.
+
+What it does, in order:
+
+1. Merge every ``*.json`` result file and deduplicate on
+   ``(source file, line, rule)``.  Findings in ``artifacts/common/**`` are
+   reported once per template that ``$ref``s them, so the raw count is several
+   times the number of real findings.
+2. Split the unique findings into *documented as expected* and *to review*.
+   A finding is expected when its Spectral JSON path matches an entry in the
+   ``suppress_schema_paths`` allowlist of the matching rule in the tooling rule
+   metadata -- the same allowlist, read from the same file, that CAMARA
+   Validation applies in API repositories.
+3. Emit annotations: one per finding to review at its native Spectral severity,
+   plus one ``notice`` per source file summarising that file's expected
+   findings.  GitHub renders at most 10 annotations per level per step, so
+   collapsing the expected baseline keeps the warning level free for findings a
+   pull request actually introduces.
+4. Print a reconciliation report -- raw versus unique counts, per-source-file
+   totals, and every finding to review -- to stdout and to the step summary, so
+   the numbers in the log, in the annotations and on the run page agree.
+
+Exit status is 0 unless the arguments themselves are unusable; the lint outcome
+is the Spectral exit status, which the workflow tracks separately.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
+from pathlib import Path
+
+import yaml
+
+# Spectral DiagnosticSeverity -> GitHub annotation level.  Mirrors OUTPUT_TYPES
+# in @stoplight/spectral-formatters' github-actions formatter, so individually
+# annotated findings keep the level they had before this script existed.
+SEVERITY_TO_LEVEL = {0: "error", 1: "warning", 2: "notice", 3: "notice"}
+
+# Where the expected findings are explained.  Referenced from the aggregate
+# annotations, which name no rule themselves.
+EXPECTED_FINDINGS_DOC = "artifacts/linting_rules/README.md"
+
+# GitHub renders at most this many annotations per level per step.
+ANNOTATION_LIMIT_PER_LEVEL = 10
+
+
+def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--findings-dir",
+        required=True,
+        type=Path,
+        help="directory holding one Spectral JSON result file per linted template",
+    )
+    parser.add_argument(
+        "--rules",
+        required=True,
+        type=Path,
+        help="tooling validation/rules/spectral-rules.yaml (source of the allowlist)",
+    )
+    parser.add_argument(
+        "--repo-root",
+        default=Path.cwd(),
+        type=Path,
+        help="repository root, used to make Spectral's absolute paths relative",
+    )
+    parser.add_argument(
+        "--status-file",
+        type=Path,
+        help="optional TSV of '<exit status>\\t<template>' lines, one per invocation",
+    )
+    parser.add_argument(
+        "--summary-file",
+        type=Path,
+        help="optional file to append the report to (normally $GITHUB_STEP_SUMMARY)",
+    )
+    return parser.parse_args(argv)
+
+
+def load_allowlist(rules_path: Path) -> dict[str, tuple[str, ...]]:
+    """Map Spectral rule code -> allowlisted schema paths from rule metadata."""
+    entries = yaml.safe_load(rules_path.read_text(encoding="utf-8")) or []
+    allowlist: dict[str, tuple[str, ...]] = {}
+    for entry in entries:
+        if not isinstance(entry, dict) or entry.get("engine") != "spectral":
+            continue
+        paths = entry.get("suppress_schema_paths") or []
+        rule = entry.get("engine_rule")
+        if rule and paths:
+            allowlist[str(rule)] = tuple(str(p) for p in paths)
+    return allowlist
+
+
+def is_expected(schema_path: str, allowed: tuple[str, ...]) -> bool:
+    """Match a schema path against an allowlist the way the post-filter does.
+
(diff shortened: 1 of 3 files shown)


rartych (approved):
LGTM

PedroDiez (approved):
LGTM

bigludo7 (approved):
LGTM

eric-murray (approved):

PR #688, opened 2026-07-30
Fix Spectral artifact finding attribution

### LINKED ISSUE #684: Artifacts Lint: Spectral step loses finding attribution and under-reports the real findings

LarryHu0217:
#### What type of PR is this?

* correction

The Artifacts Lint workflow currently passes all four templates to one Spectral invocation. Shared `$ref` targets trigger stoplightio/spectral#2640, producing 152 results with 143 missing their source and hiding six genuine findings from the reported subset.

This runs Spectral once per template and deduplicates the JSON results by file, line, and rule. The formatter reads the pinned S-313 `suppress_schema_paths` metadata, summarizes the ten documented common-library findings as two notices, and leaves the other five findings as individual warnings at their real source locations. The resulting log and annotations consistently account for all 15 unique findings.

Fixes #684

- [ ] Yes
- [x] No

The formatter tests cover source normalization, duplicate suppression, allowlist-only grouping, native severity mapping, and GitHub annotation escaping. The workflow continues to block only on error-level findings.

#### Changelog input

```
Fix Spectral source attribution and annotation accounting in the common-artifact lint workflow.
```

#### Additional documentation

The artifacts linting README now documents per-template Spectral execution, deduplication, and S-313 baseline notices.

### DIFF
--- a/.github/workflows/artifacts-lint.yml
+++ b/.github/workflows/artifacts-lint.yml
@@ -22,6 +22,8 @@
 #
 # Changelog:
 # - 2026-07-16: Initial version (yamllint + Spectral + gplint)
+# - 2026-07-30: Preserve Spectral attribution with per-template runs and
+#               collapse the documented S-313 common-library baseline.
 #
 # SEE ALSO: artifacts/linting_rules/README.md
 # =========================================================================================
@@ -69,6 +71,7 @@ jobs:
             validation/package.json
             validation/package-lock.json
             validation/.npmrc
+            validation/rules/spectral-rules.yaml
           sparse-checkout-cone-mode: false
 
       - name: Setup Python
@@ -84,6 +87,9 @@ jobs:
       - name: Install yamllint (tooling-pinned version)
         run: pip install --quiet -r .tooling/requirements.txt
 
+      - name: Test Spectral report formatter
+        run: python -B scripts/test_report_spectral_findings.py
+
       - name: Install Spectral and gplint (tooling-pinned versions)
         run: npm ci --ignore-scripts
         working-directory: .tooling/validation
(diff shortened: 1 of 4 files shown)


rartych:
Thank you for the thorough work and tests here.

I'm closing this PR in favor of #685, which implements the same core fix (run Spectral per-template, dedupe results, and collapse the documented baseline into notice annotations) while also pulling the authoritative S-313 allowlist from the tooling repo and adding the expected-findings documentation (the single-source-of-truth). 

Your formatter tests are valuable; they could be considered for inclusion in the future, if needed.

PR #698, opened 2026-08-11
Alignment of Pagination Datatype

### LINKED ISSUE #696: Inconsistent Pagination example
**Problem description**

The Pagination example in [Section 4.1.3](https://github.com/camaraproject/Commonalities/blob/main/documentation/CAMARA-API-Design-Guide.md#413-response-body) is not consistent with the definitions in CAMARA_common.yaml (assuming that CAMARA_common.yaml is taking the precedence). For example, the type (`integer`) is wrong and there is no `maximum`.

**Expected action**

Correct CAMARA-API-Design-Guide.md using the same definitions as in CAMARA_common.yaml.

**Additional context**

tlohmar:
#### What type of PR is this?

* correction

The PR aligns the Pagination Datatype definition in the example to CAMARA_common.yaml.

Fixes #696

- [ ] Yes
- [x] No

#### Changelog input

```
Alignment of the Pagination Datatype definition in the example to CAMARA_common.yaml.

```

#### Additional documentation 

This section can be blank.

```
docs

```

### DIFF
--- a/documentation/CAMARA-API-Design-Guide.md
+++ b/documentation/CAMARA-API-Design-Guide.md
@@ -645,22 +645,49 @@ ResourceList:
     pagination:
       $ref: "#/components/schemas/Pagination"
 
-Pagination:
-  type: object
-  properties:
-    page:
+    Pagination:
+      description: Pagination details helping to navigate through paged results efficiently.
+      type: object
+      properties:
+        page:
+          $ref: "#/components/schemas/Page"
+        perPage:
+          $ref: "#/components/schemas/PerPage"
+        totalCount:
+          $ref: "#/components/schemas/TotalCount"
+        totalPages:
+          $ref: "#/components/schemas/TotalPages"
+
+    Page:
       type: integer
+      format: int32
       minimum: 1
-    perPage:
+      maximum: 2147483647
+      default: 1
+      description: Current page number (1-indexed).
+      example: 1
+    PerPage:
       type: integer
+      format: int32
       minimum: 1
       maximum: 100
-    totalCount:
+      default: 20
+      description: Number of items per page.
+      example: 20
+    TotalCount:
       type: integer
+      format: int32
       minimum: 0
-    totalPages:
+      maximum: 2147483647
+      description: Total number of items matching the query, after filters applied. MAY be omitted where a full count query is prohibitively expensive.
+      example: 87
+    TotalPages:
       type: integer
+      format: int32
       minimum: 0
+      maximum: 2147483647
+      description: Total number of pages. Equals ceil(totalCount / perPage). MAY be omitted where totalCount is omitted.
+      example: 5
 ```
 
 #### 4.1.4. Response Headers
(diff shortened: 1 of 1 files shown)


eric-murray (dismissed):

documentation/CAMARA-API-Design-Guide.md
```diff
@@ -645,22 +645,49 @@ ResourceList:
     pagination:
       $ref: "#/components/schemas/Pagination"
 
-Pagination:
-  type: object
-  properties:
-    page:
+    Pagination:
```
PedroDiez: Suggest to align indentation of these schemas (Pagination, Page,...) to be at the same level as ResourceList

PedroDiez (commented):
LGTM

documentation/CAMARA-API-Design-Guide.md
tlohmar: done

PedroDiez (dismissed):
LGTM

bigludo7 (dismissed):
LGTM

rartych (dismissed):
LGTM

documentation/CAMARA-API-Design-Guide.md
```diff
@@ -646,21 +646,48 @@ ResourceList:
       $ref: "#/components/schemas/Pagination"
```
hdamker: The fix below duplicates the complete definition from CAMARA_common.yaml which should be referenced by all APIs using pagination. In this sense CAMARA_common.yaml owns now the definition of `Pagination` + `Page`/`PerPage`/`TotalCount`/`TotalPages`. The PR would resolve the drift today correctly but it will happen again. 

Also would the example here disagree with the API templates which does `$ref: "../common/CAMARA_common.yaml#/components/schemas/Pagination"` instead defining the schema locally.

In line 546 we have here in the same document already a precedent for using `../common/CAMARA_common.yaml#/...` instead of using local definitions.

And finally: the example defines the envelope, not the Pagination itself "defines a paginated response schema". 

Therefore I suggest to delete the whole local Pagination block and change the line 646 to:

```suggestion
      $ref: "../common/CAMARA_common.yaml#/components/schemas/Pagination"
```

hdamker (changes requested):
See suggestion below ... instead of adding about 27 lines I suggest to delete 13 lines. We should avoid to repeat definitions from CAMARA_common.yaml verbatim going forward, and we should encourage API authors in the examples to use refs instead of local definitions.

documentation/CAMARA-API-Design-Guide.md
```diff
@@ -646,21 +646,48 @@ ResourceList:
...
+  maximum: 2147483647
+  description: Total number of items matching the query, after filters applied. MAY be omitted where a full count query is prohibitively expensive.
+  example: 87
+TotalPages:
+  type: integer
+  format: int32
+  minimum: 0
+  maximum: 2147483647
+  description: Total number of pages. Equals ceil(totalCount / perPage). MAY be omitted where totalCount is omitted.
+  example: 5
```
hdamker: Once `pagination` refs `CAMARA_common.yaml` above, this local copy of `Pagination`/`Page`/`PerPage`/`TotalCount`/`TotalPages` should be removed — nothing points to it.

tlohmar: @hdamker @rartych Please check. I have also removed the Note above the example and linked common.yaml for the detailed schema definition.

hdamker (approved):

bigludo7 (approved):
LGTM

rartych (approved):
LGTM

eric-murray (approved):

PR #711, opened 2026-09-15
Unique Scenario names

### LINKED ISSUE #709: [Test Definitions] Warning on duplicated test scenario names
**Problem description**

The API Testing guideline only require unique tags (i.e. @<(mandatory)feature_identifier>_<...) and there is no mentioning of unique Scenario names. Seems that the verification is doing more than the Guidelines.

**Expected action**

Review and clarify API Testing guidelines wrt unique scenario names

**Additional context**

The issue was detected during review of [Dedicated Networks rc.1 snapshot](https://github.com/camaraproject/DedicatedNetworks/pull/181#issuecomment-5666389418).

tlohmar:
#### What type of PR is this?

* documentation

The PR suggests a clarification of the Test Definitions, matching the verification checks. A statement is added, that the Scenario name must be unique.

Fixes #709

- [ ] Yes
- [x] No

#### Changelog input

```
Clarifying that the Scenario names must be unique

```

#### Additional documentation 

This section can be blank.

```
docs

```


documentation/API-Testing-Guidelines.md
```diff
@@ -207,7 +207,7 @@ Background:
 
 * It is recommended to only have one When/Then block per scenario. However, in case of complex scenarios, several When/Then blocks can be concatenated in the same scenario.
 
-* Each scenario will have a tag with the scenario unique identifier. The recommended format for scenario identifier is:
+* Each scenario must have a unique `Scenario` name (or `Scenario Outline` name) and a unique tag identifier. The recommended format for scenario tag identifier is:
```
PedroDiez: Would suggest to use wording "MUST" to follow RFC2119

PedroDiez: LGTM with the update

hdamker: Not sure if the following is the best way to express it, but the requirement for unique scenario name is only within a feature. Across feature (files) scenario names can be reused. In other words: a scenario is identified by feature name + scenario name.

```suggestion
* Each scenario MUST have a unique `Scenario` name (or `Scenario Outline` name) within a feature and a unique tag identifier. The recommended format for scenario tag identifier is:
```

documentation/API-Testing-Guidelines.md
tanjadegroot, line 219:
```suggestion
Each defined error response need its own scenario with a unique scenario name per operation. Alternatively, multiple, operation specific feature files can be created as single feature files can get very large, making them difficult to maintain.

```

documentation/API-Testing-Guidelines.md
tlohmar: I preferred "must" over "MUST" to stay aligned with the rest of the document. I assume that the first version was created before CAMARA started following RFC2119.

Could make sense to use this PR as trigger and capitalize all the other keywords as well.

documentation/API-Testing-Guidelines.md
tlohmar: Isnt that clear from the structure of steps, i.e. error response is checked with a `Then` statement, meaning, each error code require an own Scenario?

Or, is that prohibiting the definition of a single `Scenario Outline`, checking for different error codes?

tlohmar:
Note, there are multiple occurrences of "must" (all small letters) in the document and this PR introduces now inconsistency with the capitalization.

documentation/API-Testing-Guidelines.md
tanjadegroot: It was just a reminder to make it explicit, but feel free to adapt or not use my proposed text.

On the Scenario Outline approach, Eric had responded as follows:

"A third approach was used by IoTDeviceManagement, which was to define the operations as different examples for a common scenario. This works well when the behaviour is largely independent of the operation (e.g. 401 UNAUTHENTICATED - see example below), but would quickly become cumbersome for operation specific behaviour.

My own preference is to have separate feature files for each operation, as single feature files can get very large, making them difficult to maintain and review.

  @esim_profile_errors_401_01_no_valid_token
  Scenario Outline: <operationId> is rejected without a valid access token
    Given the resource "/esim-profile-management/vwip/<endpoint>"
    And the header "Authorization" is set to an invalid access token
    When the request "<operationId>" is sent
    Then the response status code is 401
    And the response property "$.status" is 401
    And the response property "$.code" is "UNAUTHENTICATED"

    Examples:
      | endpoint       | operationId            |
      | download       | downloadEsimProfile    |
      | enable         | enableEsimProfile      |
      | disable        | disableEsimProfile     |
      | delete         | deleteEsimProfile      |
      | set-fallback   | setFallbackEsimProfile |
      | retrieve-status| getEsimProfileStatus   |

So I would not put the Scenario Outline in the recommendations.

